PR #38, opened 2018-11-22
Download raw without solr lookup

blekinge:
As the title says


thomasegense:
Break functionality. Download without solrlookup is required for forwards.

PR #72, opened 2020-08-24
Fix naming and add images

jesperlauridsen:
Minor bug fixes
Added images to results
Made it so you can press image to see the big version
Minor styling fixes.


src/js/src/components/SearchSingleItemComponents/SearchSingleItemFocusImage.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <div class="searchSingleItemFocusImage" @click="closeImage">
+    <img :src="image" @click="stopCloseImage">
+    <button @click="closeImage">
```
jorntx, line 4: You could try @click.stop here and see if you can do without stopCloseImage() function.

src/js/src/components/SearchSingleItemComponents/SearchSingleItemAllData.vue
```diff
@@ -2,7 +2,7 @@
   <div class="SingleEntryAllData">
     <div class="showAllButtonContainer">
       <button class="allDataButton" @click="toggleAllDataShown">
-        See all data
+        {{ allDataButtonText }}
```
jorntx, line 5: Consider using a computed property for the text toggling and then just toggle the `allDataShown` property above directly in the click handler. Gives a bit cleaner code in this case.

src/js/src/components/SearchSingleItemComponents/SearchSingleItemImages.vue
```diff
@@ -0,0 +1,105 @@
...
+      imageSrcs:[],
+      allImagesShown:false,
+      shownImagesButtonText: 'See all images',
+      showFullImage:null
+    }
+  },
+  computed: {
+  },
+  watch: { 
+    source: function() { // watch it
```
jorntx, line 62: Out of curiosity - why du you need to watch the source here?

src/js/src/components/SearchSingleItemComponents/SearchSingleItemImages.vue
```diff
@@ -0,0 +1,105 @@
...
+  <div v-if="imageSrcs.length > 0" class="singleEntryImages">
+    <div class="singleEntryImagesHeadline">
+      <p class="highlightText entryInfo imageHeadline">
+        Images:
+      </p>
+      <p> 
+        showing {{ showNumberOfPictures(imageSrcs.length) }} out of {{ imageSrcs.length }}
+      </p>
+      <button v-if="imageSrcs.length > 4" class="allImagesButton" @click="toggleAllImagesShown">
+        {{ shownImagesButtonText }}
```
jorntx, line 11: This could be a computed property instead - see my comment above.

jorntx (commented):
Looks good - have a few comments and one question. Seems to run OK but I get a HTTP 500 on the image service (with no real impact on the app) when searching for 'Toke'

src/js/src/components/SearchSingleItemComponents/SearchSingleItemImages.vue
jesperlauridsen: Because Vue tries to reuse elements when we load in now items after a new search - And i need to watch at least one of the props, to make sure we update the images when a new result is added. I noticed that the results updated, but the images didn't, so new results had old images. This solves that problem.

src/js/src/components/SearchSingleItemComponents/SearchSingleItemAllData.vue
jesperlauridsen: Fixed.

src/js/src/components/SearchSingleItemComponents/SearchSingleItemImages.vue
jesperlauridsen: Fixed.

src/js/src/components/SearchSingleItemComponents/SearchSingleItemFocusImage.vue
jesperlauridsen: Fixed.

src/js/src/components/SearchSingleItemComponents/SearchSingleItemImages.vue
jorntx: 👍

jorntx (approved):
Looks good 👌

PR #73, opened 2020-08-26
Soback 4 handle playback url

jorntx:
First stab at handling playback.

@thomasegense check that this plays nice with Tomcat and all of your playback code

@jesperlauridsen do your usual stuff but take note that this is a first stab and I have some issues with the way I handle the configs. I think they should be loaded into a real store but haven't found a clean way to do this without breaking or polluting the fancy store auto loading mechanism I put in place in my last PR.

### DIFF
--- a/src/js/src/main.js
+++ b/src/js/src/main.js
@@ -2,13 +2,25 @@ import Vue from 'vue'
 import App from './App.vue'
 import router from './router'
 import store from './store'
-
-Vue.config.productionTip = false
+import Axios from 'axios'
+import configs from './configs'
 
 import './assets/styles/main.scss'
 
+Vue.config.productionTip = false
+
+Axios.get('/services/frontend/properties/solrwaybackweb/')
+    .then(response => {
+        configs.playbackConfig.openwaybackBaseURL = response.data['openwayback.baseurl']
+        configs.playbackConfig.solrwaybackBaseURL = response.data['wayback.baseurl']
+        initializeVue()
+    })
+    .catch(error => initializeVue())
+
+function initializeVue(){
 new Vue({
   router,
   store,
   render: h => h(App)
 }).$mount('#app')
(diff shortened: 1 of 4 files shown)


src/js/src/main.js
```diff
@@ -2,13 +2,25 @@ import Vue from 'vue'
...
 Vue.config.productionTip = false
 
-import './assets/styles/main.scss'
+Axios.get('services/frontend/properties/solrwaybackweb/')
+    .then(response => {
+        configs.playbackConfig.openwaybackBaseURL = response.data['openwayback.baseurl']
+        configs.playbackConfig.solrwaybackBaseURL = response.data['wayback.baseurl']
+        initializeVue()
+    })
+    .catch(error => initializeVue())
```
jesperlauridsen, line 18: Vi har ingen fallback for hvis vi ikke kan få fat i configs? (nu aner jeg ikke, er de altid tilgængelige?)

jorntx: Good objection! Nope - no fallback. The App will start though :) I think a smart way to do this would be to make a second request later on if the first one fails. I would like to implement this when I figure out a smart way to get the configs into a real store.

jesperlauridsen: True - all the more confusing when it starts and won't work properly from a user perspective tho. But i agree, it can wait with implementation. Also, this spawns a second task. Errors - we need some sort of error handling at least. The app might start here, but it should say something like "But we cant get the paths to wayback, seaching will work, but no display of individual pages" or something of that kind. Maybe. But we need error-handling, that's my point!

jorntx: I agree

jesperlauridsen (approved):
Other than the small details mentioned (that'll be fixable later). I tested the branch, and playback seems to be working just fine. Good work!

PR #75, opened 2020-08-31
fix to query input.

jesperlauridsen:
New the input in the query textfield can always be seen. If the item is in focus you can scroll your way to the end - if it's not, and the text is longer than the text-field, you get a "..." in the end to show it's longer.


jorntx (approved):
Tested OK ;)

PR #76, opened 2020-08-31
Rough outline of search by file upload functionality

jorntx:
A very rough outline of the functionality (and sorry for the crude css but I think we should fix this in a major search box overhaul).

Take a look and see if you like where this is heading. If yes we should merge it and make some tickets for improving it.

### DIFF
--- a/src/js/src/components/SearchBox.vue
+++ b/src/js/src/components/SearchBox.vue
@@ -14,23 +14,31 @@
               id="clearSubmit"
               title="Clear search and results"
               type="button"
-              @click="clearResultsAndSearch" />     
+              @click="clearResultsAndSearch" />
     </form>
+    <div @click="showUploadFileSearch = !showUploadFileSearch">
+      <span class="searchBoxActionLink">Search with uploaded file</span>
+    </div>
     <applied-search-facets />
+    <search-upload-file v-if="showUploadFileSearch" />
   </div>
 </template> 
 
 <script>
 import { mapState, mapActions } from 'vuex'
 import AppliedSearchFacets from './AppliedSearchFacets.vue'
+import SearchUploadFile from './SearchUploadFile.vue'
 
 export default {
   components: {
-    AppliedSearchFacets
+    AppliedSearchFacets,
+    SearchUploadFile
   },
   data () {
     return {    
-      futureQuery:'' 
+      futureQuery:'',
+      showUploadFileSearch: false
+
     }
   },
   computed: {
@@ -81,9 +89,6 @@ export default {
       history.pushState({name: 'SolrWayback'}, 'SolrWayback', '/')
       this.futureQuery = ''
       this.resetSearchState()
-      //this.updateQuery('')
-      //this.clearResults()
-      //this.updateSearchAppliedFacets('')
     },
   }
 }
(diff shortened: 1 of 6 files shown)


src/js/src/components/SearchUploadFile.vue
```diff
@@ -0,0 +1,63 @@
...
+  uploadFile(){
+     //TODO spinner
+     requestService.uploadFileRequest(this.fileToUpload[0])
+        .then(response => {
+          //Search for file
+         this.updateQuery(this.createRequestQuery(response.data))
+         this.requestSearch({query:this.createRequestQuery(response.data), facets:this.searchAppliedFacets})
+        })
+        .catch(() => {
+           //TODO Call error handler - something went wrong
```
jesperlauridsen, line 51: Gotta start keeping track of our error handling places

src/js/src/components/SearchUploadFile.vue
```diff
@@ -0,0 +1,63 @@
...
+      this.fileToUpload = this.$refs.file.files
+    },
+
+  uploadFile(){
+     //TODO spinner
+     requestService.uploadFileRequest(this.fileToUpload[0])
+        .then(response => {
+          //Search for file
+         this.updateQuery(this.createRequestQuery(response.data))
+         this.requestSearch({query:this.createRequestQuery(response.data), facets:this.searchAppliedFacets})
```
jesperlauridsen, line 48: very good to update the query here aswell - the requestSearch method is being expanded in another incoming branch where you'll need the solrSettings param to be passed to the search function aswell. Just a heads up.

src/js/src/components/SearchBox.vue
```diff
@@ -14,23 +14,31 @@
...
               title="Clear search and results"
               type="button"
-              @click="clearResultsAndSearch" />     
+              @click="clearResultsAndSearch" />
     </form>
+    <div @click="showUploadFileSearch = !showUploadFileSearch">
+      <span class="searchBoxActionLink">Search with uploaded file</span>
+    </div>
     <applied-search-facets />
+    <search-upload-file v-if="showUploadFileSearch" />
```
jesperlauridsen, line 23: The placement of the button is a little weird if you start a search and place some filters. Perhaps we could find a better way to start the interaction - maybe clear the search all together and push a new state when you want to search for an uploaded file?

jesperlauridsen (approved):
All in all very nice - the functionality is solid. the UI placement could imho be worked in a little better - but theres lots of time to find out how we're gonna have all the functionality play together nicely. Approved from here! :+1:

src/js/src/components/SearchUploadFile.vue
jorntx: Yes - we need to put this task on the board

src/js/src/components/SearchUploadFile.vue
jorntx: Thanks for the heads up!

src/js/src/components/SearchBox.vue
jorntx: I totally agree here and actually I forgot about testing with filters applied :) I think you are spot on with the interaction - we should consider removing or making the search box  (appear) inactive when searching with uploaded file.

jorntx:
Thanks for the feedback - I converted your comments to tickets

PR #78, opened 2020-09-02
Soback 8 create image template

jesperlauridsen:
I decided to save some work, so I changed the SearchSingleItemImages to support both singular and multiple images. This saves a ton of work, as it handles any kind of image shown in the search result posts now - whether it's an item of the image sort (then it'll only show the one), or it's the type where the post has a ot of images (as it was before)

I think it's a good solution, reuses alot of the same code.

I also added small icons to the title of the post, showing what kind of post that's loaded.


src/js/src/components/SearchSingleItemComponents/SearchSingleItemTypes/SearchSingleItemImage.vue
```diff
@@ -0,0 +1,48 @@
...
+  </div>
+</template>
+
+<script>
+import SearchSingleItemStandardInfo from './../SearchSingleItemStandardInfo.vue'
+import SearchSingleItemAllData from './../SearchSingleItemAllData.vue'
+import SearchSingleItemImages from './../SearchSingleItemImages.vue'
+import configs from '../../../configs'
+
+//import { mapState, mapActions } from 'vuex'
```
jorntx, line 15: Unused code

src/js/src/components/SearchSingleItemComponents/SearchSingleItemTypes/SearchSingleItemImage.vue
```diff
@@ -0,0 +1,48 @@
...
+  data () {
+    return {     
+    }
+  },
+  computed: {
+  },
+  mounted () {
+  },
+  methods: {
+  }
```
jorntx, line 43: Unused code

jorntx (approved):
Solid work and very nice solution with reusing the image template 👍 

Only found som unused code - so I approved right away.

PR #80, opened 2020-09-07
Soback 6 image search continued

jesperlauridsen:
Added image search
Changed layout for image search
Made it possible to search from each images with either links_images or hash (as it is today in the old one)
Fixed bug where an updated route search was always a post search
Made so you can see images in full size


src/js/src/components/SearchBox.vue
```diff
@@ -94,14 +94,30 @@ export default {
       this.$router.history.current.query.grouping === 'true' ? this.updateSolrSettingGrouping(true) : null
       this.$router.history.current.query.imgSearch === 'true' ? this.updateSolrSettingImgSearch(true) : null
       this.$router.history.current.query.urlSearch === 'true' ? this.updateSolrSettingUrlSearch(true) : null
-      this.requestSearch({query:this.query, facets:this.searchAppliedFacets, options:this.solrSettings})
-      this.requestFacets({query:this.query, facets:this.searchAppliedFacets, options:this.solrSettings})
+      if(this.solrSettings.imgSearch === true) {
```
jorntx, line 97: Don't have to check for true with real Booleans

src/js/src/components/SearchBox.vue
```diff
@@ -94,14 +94,30 @@ export default {
...
+          console.log('firing image search')
+           this.requestImageSearch({query:this.query})
+           this.$_pushSearchHistory('SolrWayback', this.query, this.searchAppliedFacets, this.solrSettings)
+          return
+        }
+        if(this.solrSettings.urlSearch === true) {
```
jorntx, line 103: Don't have to check for true with real Booleans

src/js/src/components/SearchBox.vue
```diff
@@ -94,14 +94,30 @@ export default {
...
+          console.log('fireing url search')
+          return
+        }
```
jorntx, line 106: Consider using if, else-if and else pattern instead of if with return statements inside. Does the same in this case and easier to read.

src/js/src/components/SearchBox.vue
```diff
@@ -118,14 +134,26 @@ export default {
...
+        if(this.solrSettings.urlSearch === true) {
+          return
+        }
+        else {
+          this.requestSearch({query:this.futureQuery, facets:this.searchAppliedFacets, options:this.solrSettings})
+          this.requestFacets({query:this.futureQuery, facets:this.searchAppliedFacets, options:this.solrSettings})
+          this.$_pushSearchHistory('SolrWayback', this.query, this.searchAppliedFacets, this.solrSettings)
+          return
+        }
```
jorntx, line 156: Same comments as above.

src/js/src/components/SearchResult.vue
```diff
@@ -5,29 +5,62 @@
...
+                     :rank-number="index" />
+        </div>
       </div>
-      <div v-if="results && results !== {}" class="results">
-        <component :is="SingleEntryComponent(result.type)"
-                   v-for="(result, index) in results.docs"
-                   :key="index"
-                   :result="result"
-                   :rank-number="index" />
+      <!-- HERE COMES PICTURES // Figure out if this should be splitted out into a new component -->
```
jorntx, line 35: I agree

src/js/src/components/SearchResult.vue
```diff
@@ -5,29 +5,62 @@
...
-        <span class="tonedDownText">(total hits: {{ results.numFound.toLocaleString("en") }})</span>.
-      </span>
-      <div class="pagingContainer">
-        <button :disabled="solrSettings.offset < 20" @click="getPreviousResults()">
-          Previous 20
-        </button>
-        <button :disabled="results.cardinality ? solrSettings.offset + 20 > results.cardinality : solrSettings.offset + 20 > results.numFound" @click="getNextResults()">
-          Next 20
-        </button>
+      <!-- HERE COMES RESULTS // Figure out if this should be splitted out into a new component -->
```
jorntx, line 8: I agree

src/js/src/components/SearchResult.vue
```diff
@@ -82,6 +118,15 @@ export default {
       this.requestFacets({query:this.query, facets:this.searchAppliedFacets, options:this.solrSettings})
       this.$_pushSearchHistory('SolrWayback', this.query, this.searchAppliedFacets, this.solrSettings)
     },
+    getOffsetArray(array, number) {
+      let newArray = [...array]
```
jorntx, line 122: Nice use of spread operator here  - good for simple Arrays (and other iterable objects for that matter) 👍

src/js/src/components/SearchSingleItemComponents/SearchMasonryImage.vue
```diff
@@ -0,0 +1,99 @@
+<template>
+  <div class="columnImageContainer">
+    <div class="number">
+      #{{ row + number * rowNumber }}
+    </div>
+    <img 
+      loading="lazy"
+      :src="result.downloadUrl"
+      @click="toggleFullImage(row + number * rowNumber)">
+    <search-single-item-focus-image v-if="showFullImage === row + number * rowNumber"
```
jorntx, line 10: Out of curiosity - why is this show/hide toggling of full image based on _row + number * rowNumber_ ?

src/js/src/views/SolrWayback.vue
```diff
@@ -22,23 +22,41 @@ export default {
...
+          return
+        }
+        if(this.solrSettings.urlSearch === true) {
+          return
+        }
+        else {
+          this.requestSearch({query:this.query, facets:this.searchAppliedFacets, options:this.solrSettings})
+          this.requestFacets({query:this.query, facets:this.searchAppliedFacets, options:this.solrSettings})
+          return
+        }
```
jorntx, line 59: Same if-else comment as above

src/js/src/components/SearchBox.vue
jorntx, line 156: I see how this definitely could use some DRY'ing out now ;)

jorntx (commented):
Good stuff I particularly like the mosaic layout and the slide in/out thing on hover. 

Consider a strategy for very large result sets (unless there already is a cap on the backend).

I have a few comments to the code but nothing major.

src/js/src/components/SearchBox.vue
jesperlauridsen: :+1:

src/js/src/components/SearchBox.vue
jesperlauridsen: :+1:

src/js/src/components/SearchBox.vue
jesperlauridsen: :+1:

src/js/src/components/SearchBox.vue
jesperlauridsen: :+1:

src/js/src/components/SearchResult.vue
jesperlauridsen: \o/

src/js/src/components/SearchBox.vue
jesperlauridsen: Yea, I've made a jira JUST for that!

src/js/src/views/SolrWayback.vue
jesperlauridsen: :+1:

src/js/src/components/SearchSingleItemComponents/SearchMasonryImage.vue
jesperlauridsen: Good catch. This is not needed when the the parent only has one image to show. The focus image template is also used in t the images shown in a post, which can contain alot of images all using the same single template - so in that context i have to keep track of which image is shown. I don't have to do that here, so I've changed it.

src/js/src/components/SearchResult.vue
jesperlauridsen: Yes. I've done the splitting in the next commit.

src/js/src/components/SearchResult.vue
jesperlauridsen: Yes. I've done the splitting in the next commit.

jorntx (approved):
👍

PR #85, opened 2020-09-11
Soback 25 spinner

jesperlauridsen:
Spinner added
Loading added in the needed places
Removing loading if we get errors, so its possible to enteract with the UI again.


src/js/src/components/LoadingOverlay.vue
```diff
@@ -0,0 +1,37 @@
...
+    }
+  },
+  computed: {
+    ...mapState({
+      loading: state => state.Search.loading,
+    })
+  },
+  methods: {
+    getLine(lineOptions, number) {
+      let lines = [...lineOptions[number % 3]]
```
jorntx, line 32: Getting good at using this in the right places :)

jorntx (approved):
Nice 'spinner'! My results load så fast that I don't get to enjoy it much ;)

Seems to function OK and no comments to the code

PR #86, opened 2020-09-15
Clean up of counting structure for replaced links

tokee:
The old version over-counted replaced & not found links for rewriting. This version cleans up the counting logic and fixes the error.


thomasegense:
Accepted. Nice tests.

PR #91, opened 2020-09-21
Added to top button

jesperlauridsen:
Simple button that takes you to the top of the page if you've scrolled down. Same as in Meloar.


jorntx (approved):
Good standard implementation and I like the fact that you chose a home grown implementation here as our use case is so simple so we don't need to import a third party component with a lot of unneeded functionality.

If we find ourselves using this function in several components we will wrap the function in a component  at that time.

PR #93, opened 2020-09-23
Soback 47 alternative playback link

jorntx:
Simpel alternative playback link - only visible when content type is _html_

Appreciation in advance for helping aligning this link beside the title (without tampering too much with the existing short record DOM) 🙏


jesperlauridsen (approved):
Smooth sailing, very nice. I got nothing on this :)

PR #96, opened 2020-09-25
Soback 48

thomasegense:
Maybe Toke can review and accept this PR ?

### DIFF
--- a/src/test/java/dk/kb/netarchivesuite/solrwayback/parsers/HtmlParserUrlRewriterFromWarcTest.java
+++ b/src/test/java/dk/kb/netarchivesuite/solrwayback/parsers/HtmlParserUrlRewriterFromWarcTest.java
@@ -1,0 +1,38 @@
+package dk.kb.netarchivesuite.solrwayback.parsers;
+
+import dk.kb.netarchivesuite.solrwayback.service.dto.ArcEntry;
+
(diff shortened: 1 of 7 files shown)


src/test/java/dk/kb/netarchivesuite/solrwayback/parsers/HtmlParserUrlRewriterFromWarcTest.java
```diff
@@ -0,0 +1,38 @@
...
+     * Integration test class to parse HTML from a warc-file.
+     * Warc-files can not be in reposity so change path to a local warc-file
+     *  
+     */
+    public static void main(String []args) {
+        try {
+        String warcFile="/media/teg/1TB_SSD/solrwayback_package_3.2/indexing/warcs/denoffentlige-00000.warc";
+        long offset=2691693;
+        
+        ArcEntry arc=ArcParserFileResolver.getArcEntry(warcFile, offset);
```
tokee, line 18: This is not a unit test and if anyone else than you try to run it, it will complain.  At a minimum, make a check for the file and write a message saying that it is intended behaviour that it cannot run.

tokee (commented):
I can see why it is problematic to write a unit test for this, so with that in mind your pull request looks fine.

PR #97, opened 2020-09-25
Fixed minor mistake in check for changes. Also wrong url was used in …

jesperlauridsen:
@jorntx, code review plx
@thomasegense, please check out the branch and see if it works as you would like it to.


jorntx (approved):
Looks good. Had a hard time finding images where download raw actually gave me a better image.

PR #98, opened 2020-09-25
Should fix the issue.

jesperlauridsen:
Pretty straight forward.


jorntx (approved):
Pretty straight and it works.

PR #100, opened 2020-09-29
Fixed encoding on field searches.

jesperlauridsen:
This should fix the issue. Added encoding in the places where its needed before we fire off a search to the solr instance.

To summarize, it is when we do any kid of field search (from any of the fields in the full raw data or images)


jorntx (approved):
Pretty straight forward  - approved.

PR #102, opened 2020-09-29
Soback 55 wrong playback url

jorntx:
This ended up being a small visual bugfix branch:

1. New error message in error notification

2. Player moved above raw data to prevent it from being pushed down when viewing raw dat.

3. Removed some unused code

### DIFF
--- a/src/js/src/components/inlinePlayer/InlinePlayer.vue
+++ b/src/js/src/components/inlinePlayer/InlinePlayer.vue
@@ -33,7 +33,6 @@ export default {
     },
 
     mounted() {
-      console.log(this.result)
         this.playerInstance = videojs(this.$refs.inlinePlayerObj, this.getPlayerOptions())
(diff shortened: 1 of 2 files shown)


src/js/src/components/inlinePlayer/InlinePlayer.vue
```diff
@@ -80,18 +80,16 @@ export default {
...
-						src: this.result.url_norm,
-					}
-				]
+				  sources: this.getSource() 
         }
         return playerConf    
       },
-    
-      getFileExtension() {
-        return this.result.url_norm.match(/\.[0-9a-z]+$/i)
```
jesperlauridsen, line 93: my regex is not fantastic. Perhaps a little // explanation here about what goes on and why this is needed?

jesperlauridsen (approved):
Very nice. I don't see any issues here!

PR #106, opened 2020-10-05
Added file extension to player format error notification

jorntx:
Once again the format is stated in our error notification message if the player throws an 'No compatible source was found for this media' error.


jesperlauridsen (approved):
Looks good. I was wondering why that was removed in the last PR - just assumed it was for a reason. I like that it's back.

PR #108, opened 2020-10-06
Soback 82 navigation bug

jesperlauridsen:
This works on localhost:8080 now.

Added a /search route that allows params - and all search routes now lead to /search/:query.

'/' allows no querys anymore, and only serves as the main page. When you search you go to /search afterwards.

Small change to not display the query in the search result anymore - as requested by PO.

I've done some rigorous testing - but feel free to give it a whirl too.

### DIFF
--- a/src/js/src/App.vue
+++ b/src/js/src/App.vue
@@ -1,6 +1,7 @@
 <template>
   <div id="app">
-    <router-view />
+    <!-- En enforce component SolrWayback update when moving from '/' to '/search' -->
+    <router-view :key="$route.path" />
   </div>
 </template>
 <script>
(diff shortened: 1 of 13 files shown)


src/js/src/App.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <div id="app">
-    <router-view />
+    <!-- En enforce component SolrWayback update when moving from '/' to '/search' -->
+    <router-view :key="$route.path" />
```
jorntx, line 4: This will trigger complete re-rendering and rerun of life-cycle hooks when routing - but I guess that this is actually what you want :)

jorntx (approved):
Looks good and tested out fine.

PR #109, opened 2020-10-06
minor fix to some text-strings

jesperlauridsen:
Fixed paging button to not show when results are on the limit (ie dont show button when you're on 100 of 100).


jorntx (approved):
That did the trick.

PR #110, opened 2020-10-06
Soback 57 about text on frontpage

jesperlauridsen:
Added about container.
Added getting text from the the service.
Added button to show/hide the text.


jorntx (approved):
Code is fine but for some reason the backend service keeps hitting mig with a 500 error.

Tried to rebuild and redeploy/restart Tomcat and dev-server but no dice:(

thomasegense (approved):
Deployed it, looks good.

PR #122, opened 2020-10-28
Soback 102 103 lazyload binary streaming

thomasegense:
Could use your eyes on this PR. 
1) Refactoring of warc-parser. Loading header info in new method and keeping the file-pointer so the binary can be loaded
2)Option to load the binary byte[] when loading a new ArcEntry
3)Lazyload method on ArcEntry to return a BufferedInputStream wrapping the binaries.
This is 4 different implementations in warc-parsering ( Warc/Arc combined with no-zip/Zip)
4) Unittest for the Lazyload method, again for 4 different unittests for each combo in 3)

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/ArcParser.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/ArcParser.java
@@ -32,19 +32,172 @@ public class ArcParser extends  ArcWarcFileParserAbstract{
    *MicrosoftOfficeWebServer: 5.0_Pub
    *X-Powered-By: ASP.NET
    */
-  public static ArcEntry getArcEntry(String arcFilePath, long arcEntryPosition) throws Exception {
+  public static ArcEntry getArcEntry(String arcFilePath, long arcEntryPosition, boolean loadBinary) throws Exception {
     RandomAccessFile raf= null ;
     try{
 
       if (arcFilePath.endsWith(".gz")){ //It is zipped
-        return getArcEntryZipped(arcFilePath, arcEntryPosition);                       
+        return getArcEntryZipped(arcFilePath, arcEntryPosition, loadBinary);                       
       }
+      else {
+          return getArcEntryNotZipped(arcFilePath, arcEntryPosition, loadBinary);
+      }      
+    }
+    catch(Exception e){
+      throw e;
+    }
+    finally{
+      if (raf!= null){
+        raf.close();
+      }
+    }
+  }
+
+  
+  public static ArcEntry getArcEntryNotZipped(String arcFilePath, long arcEntryPosition, boolean loadBinary) throws Exception {
+      RandomAccessFile raf= null ;
       ArcEntry arcEntry = new ArcEntry();
-      StringBuffer headerLinesBuffer = new StringBuffer();
+      arcEntry.setSourceFilePath(arcFilePath);
+      arcEntry.setOffset(arcEntryPosition);
+      
+      try{        
+      
+        raf = new RandomAccessFile(new File(arcFilePath), "r");
+        raf.seek(arcEntryPosition);
 
+        loadArcHeaderNotZipped(raf, arcEntry);
+        
+        long binarySize=arcEntry.getBinaryArraySize();
+        
+        if (loadBinary) {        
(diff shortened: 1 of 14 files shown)


src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/ArcParser.java
```diff
@@ -238,6 +240,34 @@ private static void loadArcHeaderNotZipped(RandomAccessFile raf, ArcEntry arcEnt
...
+          
+          RandomAccessFile raf = new RandomAccessFile(new File(arcFilePath), "r");
+          raf.seek(arcEntryPosition);          
+
+          // log.info("file is zipped:"+arcFilePath);
+          InputStream is = Channels.newInputStream(raf.getChannel());                           
+          GZIPInputStream zipStream = new GZIPInputStream(is);              
+          BufferedInputStream  bis= new BufferedInputStream(zipStream);
+
+          loadArcHeaderZipped(bis, arcEntry);
```
tokee, line 256: I cannot figure out why there are different parsers for zipped and non-zipped headers. They both operate on an unzipped `InputStream`!?

src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/ArcParser.java
```diff
@@ -32,19 +32,173 @@
...
+      RandomAccessFile raf= null ;
       ArcEntry arcEntry = new ArcEntry();
-      StringBuffer headerLinesBuffer = new StringBuffer();
+      arcEntry.setFormat(ArcEntry.FORMAT.ARC);
+      arcEntry.setSourceFilePath(arcFilePath);
+      arcEntry.setOffset(arcEntryPosition);
+      
+      try{        
+      
+        raf = new RandomAccessFile(new File(arcFilePath), "r");
```
tokee, line 66: Consider using the Java 1.7 `Closeable` feature and write
```
    try (RandomAccessFile raf = new RandomAccessFile(new File(arcFilePath), "r")){
```
instead of checking if `raf`has been created in the `finally` block.

src/main/java/dk/kb/netarchivesuite/solrwayback/service/dto/ArcEntry.java
```diff
@@ -11,14 +13,31 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.sun.tools.javac.api.Formattable;
```
tokee, line 16: This is not used and it is a Sun/Oracle-specific class, for it makes it not work under OpenJDK.

src/test/java/dk/kb/netarchivesuite/solrwayback/ArcGzParserTest.java
```diff
@@ -37,5 +38,23 @@ public void testArcGzParserImage() throws Exception {
         assertEquals(1662,arcEntry.getBinary().length); //Actually loaded in binary
     }
     
-       
+ 
+    @Test
+    public void testLazyLoadBinary() throws Exception {
```
tokee, line 43: Thank you for this!

tokee (changes requested):
The only blocker is the unused import in `ArcEntry`. The rest is optional.

src/main/java/dk/kb/netarchivesuite/solrwayback/service/dto/ArcEntry.java
thomasegense: Bummer! Huge mistake. It is not even used

src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/ArcParser.java
thomasegense: HHMM. Good question, not sure it is correct. But this must be another refactorng PR. It was not part of this PR.

src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/ArcParser.java
thomasegense: Fixed in both ArcParser and WarcParser

PR #123, opened 2020-10-28
Switch to stream-based buffering for WARC export

tokee:
`StreamingSolrWarcExportBufferedInputStream` now holds a buffer of `InputStream`s instead of `byte[]` and uses a buffer for its `read(b, off, len)`-method, which should be faster.

There is no unit-test, but @thomasegense has a local test that he can run.


src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/ArcParser.java
```diff
@@ -235,7 +237,7 @@ private static void loadArcHeaderNotZipped(RandomAccessFile raf, ArcEntry arcEnt
   }
   
 
-  public static BufferedInputStream lazyLoadBinary(String arcFilePath, long arcEntryPosition) throws Exception{
+  public static BufferedInputStream lazyLoadBinary(String arcFilePath, long arcEntryPosition, long binarySize) throws Exception{
```
thomasegense, line 239: Dont think it is necessary to add binarySize to the API.
Both the warc and arc parser re-parse the header to the get file-pointer to the correct position. And in this 
parsing there binarySize is calculated again.

thomasegense:
Only single a review comment. Fix if you agree.
Looked at the new  StreamingSolrWarcExportBufferedInputStream but take your words for it is ok :)

src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
```diff
@@ -253,7 +254,7 @@ public static ArcEntry getWarcEntryZipped(String warcFilePath, long warcEntryPos
   }
 
     
-    public static BufferedInputStream lazyLoadBinary(String arcFilePath, long arcEntryPosition) throws Exception{
+    public static BufferedInputStream lazyLoadBinary(String arcFilePath, long arcEntryPosition, long binarySize) throws Exception{
```
thomasegense, line 257: Dont think it is necessary to add binarySize to the API.
Both the warc and arc parser re-parse the header to the get file-pointer to the correct position. And in this
parsing there binarySize is calculated again.
Sp same issue in lazyloadBinary method in ArcParser.

thomasegense (changes requested):

src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/ArcParser.java
tokee: Nice catch. Fixed.

PR #124, opened 2020-10-30
Soback 107: Add gzip support for WARC export

tokee:
This pull request makes it optional if exported WARC records should be as-is or gzipped.


thomasegense (approved):
Approved - further testing will be on newfrontend branch, but seems to work.

PR #125, opened 2020-11-02
Soback 107 smurf implementation

jorntx:
Smurf functionality added to toolbox. 

Egense added as co-reviewer to test out that all is good in Tomcat as this PR tampers with routing.


thomasegense (changes requested):
Needs to stay in the toolbox at the cost of navigation.

thomasegense (approved):
Working inside toolbox. OK

src/js/src/components/ngrams/chartsCore/chartConfigs/chartOptionsCore.js
```diff
@@ -0,0 +1,201 @@
...
+        }
+      }]
+    }
+  },
+
+  /**
+  * Generate random color for lines
+  * Kudos - https://stackoverflow.com/questions/10014271/generate-random-color-distinguishable-to-humans 
+  */
+  getChartLineColor(salt){
```
jesperlauridsen, line 117: Very cool!

src/js/src/components/ngrams/chartsCore/chartEngines/LineChart.js
```diff
@@ -0,0 +1,28 @@
...
+    },
+    options: {
+      type: Object,
+      default: null
+    }
+  },
+  mounted () {
+    // this.chartData is created in the mixin.
+    // If you want to pass options please create a local options object
+    this.renderChart(this.chartData,  this.options)
```
jesperlauridsen, line 21: This seems funky to me somehow. You never really create a new chart (like let chart = new Chart()) - but it somehow works with this vue-chartjs magic. Very funky, imho

src/js/src/store/modules/ngram.store.js
```diff
@@ -0,0 +1,116 @@
+import { requestService } from '../../services/RequestService'
+
+
+// Global ngram state.
+const initialState = () => ({
+  query: '',
+  // cache for current query in case server rejects it
+  attemptedQuery:'',
```
jesperlauridsen, line 8: You don't use this attemptedQuery anywhere? At least I can't seem to find it...

src/js/src/components/ngrams/netarchive/SearchBox.vue
```diff
@@ -0,0 +1,75 @@
...
+ 
+  data () {
+    return {    
+        searchQuery:''
+    }
+  },
+  
+  computed: {
+    ...mapState({
+     query: state => state.Ngram.query,
```
jesperlauridsen, line 37: You also don't use the query from the store here it seems. You don't have the method for updating it set in your methods picked from the Ngram store in mapActions.

src/js/src/components/ngrams/chartsCore/chartEngines/LineChart.js
jorntx: Very valid point and one of the drawbacks of using wrappers:(

I think we perhaps should have a future discussion about the usage of wrappers vs 'the raw product code'. 

The logic here is that the LineChart component is an instance of ChartJs hence the _this.renderChart_

jesperlauridsen: But isn't one of the drawbacks here, that you can't get the chart object (which you seem to need in your onClick function that's causing you problems?

src/js/src/store/modules/ngram.store.js
jorntx: Yes, this is dead code - good catch 👍

src/js/src/store/modules/ngram.store.js
```diff
@@ -0,0 +1,116 @@
...
+  results: {},
```
jesperlauridsen, line 9: state.results isn't really used for anything significant. You put the current result in it after a successfull search, but use the state.datasets for drawing the chart. The only real use, as far as i can see, is in SearchBox.vue where you use it to check if the "reset"-button should be shown - but you can just look at the state.dataset variable instead. I think it should be removed to nice up the code :-)

src/js/src/components/ToolboxComponents/NgramNetarchive.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+  <div class="linkGraphContainer">
+    <h2 class="toolboxHeadline">
+      Ngram Netarchive
+    </h2>
+    <div class="contentContainer">
+      <transition name="loading-overlay">
+        <loading-overlay />
+      </transition>
+      <notifications />
```
jesperlauridsen, line 10: Why are we needing the notifications here again? Just curious...

src/js/src/components/ngrams/chartsCore/chartConfigs/chartOptionsCore.js
jorntx: :)

src/js/src/components/ngrams/chartsCore/chartEngines/LineChart.js
jorntx: I have access to the chart object here in LineChart no problem (also stated in the docs) but if I tamper with _this.option_ here even as data/computed property(!!) all hell breaks loose on first search. After first search all is fine.

src/js/src/components/ngrams/netarchive/SearchBox.vue
jorntx: 👍

src/js/src/store/modules/ngram.store.js
jorntx: Another good catch!

src/js/src/components/ToolboxComponents/NgramNetarchive.vue
jorntx: You are on 🔥 today! Leftover from the refactor in toolbox we can rely on the global one:)

src/js/src/components/ngrams/chartsCore/chartEngines/LineChart.js
jesperlauridsen: Hmmm. It sounds... intriguing! I hope you find your solution to this at some point :pray:

jesperlauridsen (approved):
Looks very good. Good to get this into the application aswell. Well done! The only few things I found are just dead code, that could be removed, I think. Tested the branch out too - apart from the error you're already aware off, it works very nicely.

PR #128, opened 2020-11-03
fixed deselect bug by checking for presence of chartObj, removed cons…

jorntx:
Pretty straightforward one-liner:)

Review just to double check that existing chart functionality doesn't break.


jesperlauridsen (approved):
Looks good!

PR #129, opened 2020-11-03
Soback 113 prevent duplicate chart entries

jorntx:
Added simple check for duplicates as requested by @tokee.


jesperlauridsen (approved):
Very nice. I like the extra mile with the suggested query :D

PR #132, opened 2020-11-05
styling fix

jesperlauridsen:
Not much code to it really.


jorntx (approved):
Good job brushing up. In a future release I think we should work with graying out/visually down playing the primary search box when using the uploader.

PR #134, opened 2020-11-05
Fix to both issues in the JIRA

jesperlauridsen:
ordered the document fields alpabetically
Removed score from each post. *cries*

@jorntx added for code review
@thomasegense added for functionality test.

### DIFF
--- a/src/js/src/components/searchSingleItemComponents/SearchSingleItemAllData.vue
+++ b/src/js/src/components/searchSingleItemComponents/SearchSingleItemAllData.vue
@@ -105,10 +105,12 @@ export default {
       }
     },
     orderResult(result) {
+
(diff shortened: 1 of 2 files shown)


thomasegense (changes requested):
sort is ok and score is removed from the field list. It is not a data field, it is a fictive field so it should be removed.

But the score value should not be removed from the result list.

jorntx (approved):
Code wise this is fine - approved! 

Be aware that some day you will be asked to sort tens of thousands of entries on the client and when that day comes you should stick with the _for
loop_ but until that day _forEach_ wins on clarity. 

Thought the score was useful...

thomasegense:
OK

PR #135, opened 2020-11-05
Fixed.

jesperlauridsen:
I could only find the one place on the facet count where it should be applied. Please do tell me if you find any other places to do it - its really easy.

@thomasegense added to check he's happy
@jorntx added for accept and merge because it's a REALLY small fix.


thomasegense (approved):
OK

jorntx (approved):
Looks fine to me but I don't have a very large dataset. On another note - we will have to revisit this when implementing i18n

PR #136, opened 2020-11-05
Removed paging and facets when no hits are returned from the respective services.

jesperlauridsen:
removed paging from post results when no results are returned (numFound/cardinality)

Removed facets when none are returned (check of the facets.facet_fields object. If we ever decide that OTHER facets should be put into play from any of the other arrays, this needs to be updated)

@jorntx for code review
@thomasegense for functionality

### DIFF
--- a/src/js/src/components/SearchFacetOptions.vue
+++ b/src/js/src/components/SearchFacetOptions.vue
@@ -51,16 +51,12 @@ export default {
     },
     checkForFacets(facets) {
     //we test if the variable exists first - can cause problems if it's not set yet.
-    if(facets) {
       let fate = false
-      Object.keys(facets).forEach(function(item) {
-        facets[item].length !== 0 ? fate = true : null
-      })
-      return fate
-      }
-      else {
-        return false
-      }
+      if(facets) { 
+        Object.keys(facets).forEach(function(item) {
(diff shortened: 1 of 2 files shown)


thomasegense (approved):
OK

src/js/src/components/SearchFacetOptions.vue
```diff
@@ -46,6 +48,19 @@ export default {
...
+    if(facets) {
+      let fate = false
+      Object.keys(facets).forEach(function(item) {
+        facets[item].length !== 0 ? fate = true : null
+      })
+      return fate
+      }
+      else {
+        return false
+      }
```
jorntx, line 63: This can be optimised: `
 let fate = false      
if(facets) {
      Object.keys(facets).forEach(function(item) {
        facets[item].length !== 0 ? fate = true : null
      })
     }
       return fate
 `

jorntx (commented):
Good fix. One minor improvement suggested.

src/js/src/components/SearchFacetOptions.vue
jorntx: Don't now where this ended but the branch is good to go so if you wan't to save the suggested improvement for later give me a ping and I will approve and merge in.

jorntx (approved):
Good to go!

PR #138, opened 2020-11-10
start year for visualizations is now configurable in Tomcat properties

jorntx:
Made start year configurable from server properties.

@jesperlauridsen not much to it but in a future release we have to look at how we handle properties. They are somewhat scattered throughout the project, have inconsistent naming and the serverside properties are loaded into the project in a clumsy way.

I will make a jira for this.


jesperlauridsen (approved):
Looks good, tested out and it works aswell.

jorntx:
Is this good to go @thomasegense?

PR #144, opened 2020-11-17
Soback 130 maximum time difference

jorntx:
Second stab at this function - thanks for the feedback!


jesperlauridsen (approved):
My dataset is too small to really give an impression of what this actually does - but it looks very very nice, so i'll approve it and hopefully we can test it on a larger corpus on the test-system.

PR #145, opened 2020-11-18
Soback 71 linkgraph continued

jesperlauridsen:
Max node set to 40
Made sure we reset the nessecary variables when doing a new search from the graph
Made sure the domain is updated


thomasegense (approved):
Approved!

PR #146, opened 2020-11-18
Added two new export options.

jesperlauridsen:
Quite simple really, had to do a little fiddling with the css and opted for 2 new methods where you could have chosen to do the 2 previous ones with args, but this seems simpler.

As always, @thomasegense for functionality and @jorntx for code review.

### DIFF
--- a/src/js/src/assets/styles/results.scss
+++ b/src/js/src/assets/styles/results.scss
@@ -782,7 +782,7 @@ margin-left:40px;
 }
 
 .downloadSearchResultDropdown:hover {
-  height:160px;
+  height:180px;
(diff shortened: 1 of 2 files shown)


jorntx (approved):
Looks good and works!

thomasegense (changes requested):
Logic wrong when showing options:
The parameters in solrwaybackweb.properties
allow.export.warc=false
allow.export.csv=true

If cvs is true : allow the single export (csv)
It warc is true: allow all 4 warc exports

If both true (allow all 5)
If both false (allow none)

thomasegense (approved):

PR #147, opened 2020-11-19
SOBACK-115 & SOBACK-116 + minor UI improvement

jesperlauridsen:
Added error message when we reach 500 or more images

Added error when backend doesn't reply properly

Made it so when you press enter in any of the longitude/latitude/radius/query input fields, we fire a search. Was tedious work having to press the button each time >.<

@thomasegense for review of functionality
@jorntx for code review (not much to look at tho, should be quick)

### DIFF
--- a/src/js/src/store/modules/notifier.store.js
+++ b/src/js/src/store/modules/notifier.store.js
@@ -24,12 +24,7 @@
   const mutations = {
    
     setNotification(state, notification) {
-      if(notification === undefined) {
-        state.notifications.push(notification)
-      }
-      else {
-        state.notifications.push(notification)
-      }
+      state.notifications.push(notification)
     },
 
     dismissNotification(state, notification) {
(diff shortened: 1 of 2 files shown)


thomasegense (approved):
OK

src/js/src/store/modules/notifier.store.js
```diff
@@ -24,7 +24,12 @@
   const mutations = {
    
     setNotification(state, notification) {
-      state.notifications.push(notification)
+      if(notification === undefined) {
+        state.notifications.push(notification)
+      }
+      else {
+        state.notifications.push(notification)
+      }
```
jorntx, line 32: What is the purpose if this? 🤔

jorntx (approved):
Works as it should - good work!

Only found a small code thing I need an explanation for.

jorntx (approved):
Let's go!

PR #148, opened 2020-11-19
Searching for tags (ngram)

jorntx:
First stab at this function built into the existing chart.

@jesperlauridsen for code
@thomasegense for sanity check of tags search function

As agreed the regex check will be implemented later.

### DIFF
--- a/src/js/src/assets/styles/charts.scss
+++ b/src/js/src/assets/styles/charts.scss
@@ -92,7 +92,7 @@
 
 .searchTypeContainer {
   box-sizing: border-box;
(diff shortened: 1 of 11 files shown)


thomasegense (changes requested):
1) export data link is split over 2 lines
2) Text changes:
"Text in HTML-pages"
"HTML-tags in HTML-pages"

thomasegense (approved):
OK from me

jesperlauridsen (approved):
I apparently did the wrong branch first - So i tested this aswell, and it looks really good.

PR #149, opened 2020-11-20
Soback 134 regex check for unwanted chars in tags query

jorntx:
Small PR which includes highlight of bad queries in Ngram searches for tags.

@jesperlauridsen review https://github.com/netarchivesuite/solrwayback/pull/148 before this one. Will make merging easier.

@thomasegense you are on for functionality.

### DIFF
--- a/src/js/src/assets/styles/charts.scss
+++ b/src/js/src/assets/styles/charts.scss
@@ -70,9 +70,12 @@
 }
 
 .exportModalTrigger{
-  margin: 10px 330px 10px 0;
-  text-decoration: underline;
+  //margin: 10px 330px 10px 0;
+  border-bottom: 1px solid;
   cursor:pointer;
+  color: var(--main-highlight-color);
+background-color: transparent;
+  
 }
 
 .exportboxContent h1{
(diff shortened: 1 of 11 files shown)


thomasegense (approved):
It is working
Btw. a lot of files for such a small change?
(didnt review files)

tokee:
Since writing `<table>` when asked to write a tag is such an obvious action, what about simply removing `<` and `>` if they are at the beginning and the end of the input?

jorntx:
> Since writing `<table>` when asked to write a tag is such an obvious action, what about simply removing `<` and `>` if they are at the beginning and the end of the input?

I agree @tokee (we could do this behind the scenes) but @thomasegense has stated that it should be non intrusive (don't mess with the query) and the user should be allowed to submit the query in all cases. Maybe there is a use case behind this stance @thomasegense ?

thomasegense:
No, not in this case. It is not the same as writing an URL or domain correct. Stripping the <> is fine with me, up the GUI guys to decide.

jorntx:
If there is no obvious reason to let the query pass with <> I will refactor this PR to strip them (if present) before querying the service.

jorntx:
Fixed.

- Toned down color and language in warning about <> 
- Moved warning above searchbox to prevent autofill from blocking view
- <> are now stripped before submitting query

jesperlauridsen (approved):
This is very neat! Good work - Search workes nicely, the interaction pattern is solid and it all seems to work very smooth. The error message when you type in >< is heavy - you certainly cant miss it as a user, so thats good.

I couldn't find any problems in the code either, very smooth.

PR #152, opened 2020-11-24
Soback 129 linebreak in query textarea

jesperlauridsen:
So, this should make it possible to make linebreaks by pressing SHIFT+Enter in the search-field, and search *should* work when you do so. This is gonna need some rigorous testing to make sure nothing is broken. As far as I've seen, it works as it's supposed to.  You should now be able to:

use linebreaks (by SHIFT+Enter)
Search with said query
And c/p the URL into a new browserwindow and still get the linebreaks + results when you open a new site with the URL.
And the textarea should have a size accordingly to the content of the query the whole time.

@tokee for rigorous testing
@jorntx for code review

### DIFF
--- a/src/js/src/components/SearchBox.vue
+++ b/src/js/src/components/SearchBox.vue
@@ -14,7 +14,7 @@
                   ? decideActiveClassesForQueryBox()
                   : ''"
                 :placeholder="solrSettings.urlSearch ? 'Enter search url' : 'Enter search term'"
-                @keydown.enter.prevent="launchNewSearch()"
+                @keydown.enter="checkKeyPresses()"
                 @input="getSizeOfTextArea()" />
       <button type="button" class="searchGuidelinesButton" @click.prevent="openSelectedModal('guidelines')">
         ?
@@ -134,6 +134,8 @@ export default {
         //If this is resulted by a backbutton going to the first time the user landed on the page, we want a clean slate.
         this.resetSearchState()
       }
+      document.getElementById('query').value = this.futureQuery
+      this.getSizeOfTextArea()
   },
   
   methods: {
@@ -152,10 +154,12 @@ export default {
       updateCurrentModal:'updateCurrentModal'
     }),
     getSizeOfTextArea() {
-      console.log('we resize')
       let textarea = document.getElementById('query')
       textarea.style.height = '1px'
       textarea.style.height = (textarea.scrollHeight) + 'px'
+    },
+    checkKeyPresses() {
+      !event.shiftKey ? (event.preventDefault(),this.launchNewSearch()) : null
     },
     selectSearchMethod(selected) {
       if(selected === 'imgSearch') {
--- a/src/js/src/components/SearchBox.vue
+++ b/src/js/src/components/SearchBox.vue
@@ -7,7 +7,7 @@
       </transition>
       <textarea id="query"
                 ref="query"
-                v-model="futureQuery"
+                v-model="decodedFutureQuery"
                 type="text"
                 rows="1"
                 autofocus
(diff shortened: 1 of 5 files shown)


src/js/src/components/SearchBox.vue
```diff
@@ -135,6 +135,8 @@ export default {
         //If this is resulted by a backbutton going to the first time the user landed on the page, we want a clean slate.
         this.resetSearchState()
       }
+      document.getElementById('query').value = this.futureQuery
+      this.$_getSizeOfTextArea('query')
```
jorntx, line 139: If this can be solved with this.$refs it would be preferable. It would do the same but be more consistent with Vue

src/js/src/components/SearchBox.vue
```diff
@@ -152,6 +154,9 @@ export default {
       updateShowModal:'updateShowModal',
       updateCurrentModal:'updateCurrentModal'
     }),
+    checkKeyPresses() {
+      !event.shiftKey ? (event.preventDefault(),this.launchNewSearch()) : null
+    },
```
jorntx, line 159: I guess you utilised the natural power of textarea to support line breaks but the threw a 'shift button check' in there as the enter button is already hijacked for submit - very clever.

I now why we changed the input to textarea and all is fine with this app but in other applications we would have to thread carefully with this approach is this can wreak havoc on accessibility.

jorntx (approved):
It tested out fine on my machine and I only found one trivial thing in the code so I will approve.

tokee (changes requested):
When entering the multi line query
```
hest_
ged
```
where `_` is space, faceting works fine. Entering the query
```
hest
ged
```
causes facets to fail.

thomasegense:
Maybe @tokee   finding with missing facets is related to the "regression?" error that facets are also not loaded it you start with empty result and trigger and URL search.

tokee:
Relevant logs from Solr, searching for
```
frø
frøer
```

```
2020-11-24 12:07:42.759 INFO  (qtp2134607032-56) [c:small s:shard1 r:core_node2 x:small_shard1_replica_n1] o.a.s.c.S.Request [small_shard1_replica_n1]  webapp=/solr path=/select params={f.crawl_year.facet.sort=index&facet.field=domain&facet.field=content_type_norm&facet.field=type&facet.field=crawl_year&facet.field=status_code&facet.field=public_suffix&hl=off&indent=true&fl=id&q.op=AND&fq=record_type:response+OR+record_type:arc&rows=0&version=2.2&f.crawl_year.facet.limit=100&q=frøfrøer&wt=json&facet=true} hits=0 status=0 QTime=2
2020-11-24 12:07:42.803 INFO  (qtp2134607032-23) [c:small s:shard1 r:core_node2 x:small_shard1_replica_n1] o.a.s.c.S.Request [small_shard1_replica_n1]  webapp=/solr path=/select params={q=frø%0afrøer&hl=on&indent=true&fl=id,score,title,hash,source_file_path,source_file_offset,url,url_norm,wayback_date,domain,content_type,crawl_date,content_type_norm,type&start=0&q.op=AND&fq=record_type:response+OR+record_type:arc&rows=20&wt=json&facet=false&version=2.2} hits=2 status=0 QTime=49
```
as can be see, the `q` param for the first request (the facets) is `frøfrøer`  while it is `frø%0afrøer` for the second request.

tokee (changes requested):
The facets works not (and the `%0a` can be seen in the Solr log), but the changes seems to have introduced a regression error with paging: If I search for `killinger katte` and press **Next 20**, the query changes to `killinger%20katte` and no result is shown.

jorntx (approved):
Looks good to me 👍

tokee (approved):
Looks good now, thanks.

PR #153, opened 2020-11-26
Soback 119 query checker

jorntx:
Added @tokee's regex query checker scripts to the project and made a nice box to the left of the search box with all the cool hints.

As agreed with @jesperlauridsen and @thomasegense I have not touched the color of the query in the search box in this first edition. We are keeping this light.

Put @tokee and @thomasegense on for functionality check. It would be especially nice if you @tokee would check if your scripts actually do what you expect. My rudimentary tests checked out OK.

@jesperlauridsen as usual you are on for the whole shabang.

### DIFF
--- a/src/js/src/mixins/SearchUtils.js
+++ b/src/js/src/mixins/SearchUtils.js
@@ -110,6 +110,7 @@ export default {
     * https://jsfiddle.net/bsc6zkxy/12/   
     **/  
     $_checkQueryForBadSyntax(q) {
+        q = q.replace('\n', ' ')
         let responses = []
         if (((q.includes(' AND ')  || q.includes(' && ')) &&
             (q.includes(' OR ') || q.includes(' || '))) && 
--- a/src/js/src/mixins/SearchUtils.js
+++ b/src/js/src/mixins/SearchUtils.js
@@ -100,6 +100,61 @@ export default {
       else {
         this.deliverSearchRequest(futureQuery, updateHistory, pagnation)
       }
-    }
+    },
+
+    /*
+    * Regex query checker
+    * Kudos to tokee
+    * https://github.com/netarchivesuite/solrwayback/issues/41#issuecomment-724571414
+    * https://jsfiddle.net/bsc6zkxy/12/   
+    **/  
+    $_checkQueryForBadSyntax(q) {
+        let responses = []
+        if (((q.includes(' AND ')  || q.includes(' && ')) &&
+            (q.includes(' OR ') || q.includes(' || '))) && 
+            !q.includes('(')) {
+          responses.push('Ambiguous AND/OR. Consider adding clarifying parentheses')
+        }
+        
+        if (q.match(/ (and|or|not) /)) {
+          responses.push('Possible faulty boolean. Booleans must be uppercase - ' +  q.replace(' and ', ' AND ').replace(' or ', ' OR ').replace(' not ', ' NOT '))
+        }
+
+        if (q.match(/[`‘’''„“‟”❝❞]/)) {
+          responses.push('Smart quotes. When quoting, use simple quote signs " - ' +  q.replace(/[`‘’'„“‟”❝❞]/ + '"'))
+        }
+
+        if (q.match(/[^:"]*:[^:" ]*[^\\]:[^" ]( ?.*)$/)) {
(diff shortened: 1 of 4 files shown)


thomasegense (approved):
OK

tokee (approved):
Looks good to me. We need to checkt it together with #152 as the linebreaks might mess up the regexps. A simple replace of newline with space should fix that).

Escaping is not handled properly when searching (e.g. `url_norm:http\://kat*` is valid but fails), but it think that is an independent issue.

thomasegense:
> Looks good to me. We need to checkt it together with #152 as the linebreaks might mess up the regexps. A simple replace of newline with space should fix that).
> 
> Escaping is not handled properly when searching (e.g. `url_norm:http\://kat*` is valid but fails), but it think that is an independent issue.

This a serious regression issue for much functionality if true?

This query seems to work in the mas
url_norm:http\://statsbib*

Try on: https://solrwb-pligt.kb.dk:4000/solrwayback/

jorntx:
> > Looks good to me. We need to checkt it together with #152 as the linebreaks might mess up the regexps. A simple replace of newline with space should fix that).
> > Escaping is not handled properly when searching (e.g. `url_norm:http\://kat*` is valid but fails), but it think that is an independent issue.
> 
> This a serious regression issue for much functionality if true.
> 
> This query seems to work as expected on master branch.
> `url_norm:http\://statsbib*`
> 
> Try on: https://solrwb-pligt.kb.dk:4000/solrwayback/

Weird...

This PR does nothing to the query. It only provides hints. So the regression must have happened earlier.

jorntx:
> Looks good to me. We need to checkt it together with #152 as the linebreaks might mess up the regexps. A simple replace of newline with space should fix that).
> 
>

Good point about checking it with #152 - @jesperlauridsen we should keep an eye on this when #152 is ready.

jorntx:
Merged #152 into this branch - time re-check @tokee

src/js/src/mixins/SearchUtils.js
```diff
@@ -101,6 +101,61 @@ export default {
...
+    },
+
+    /*
+    * Regex query checker
+    * Kudos to tokee
+    * https://github.com/netarchivesuite/solrwayback/issues/41#issuecomment-724571414
+    * https://jsfiddle.net/bsc6zkxy/12/   
+    **/  
+    $_checkQueryForBadSyntax(q) {
+        let responses = []
```
tokee, line 113: A simple `q = q.replace('\n', ' ');` should ensure that multi-line queries are also checked for syntactic problems.

tokee (changes requested):
As feared/expected, multi-line queries are not handled properly by the regexps. The easy fix is to make them not multi-line.

tokee (approved):
Works well, go ahead!

src/js/src/mixins/SearchUtils.js
jesperlauridsen: Codescanner seems to have a point:
Check warning on line 113 in src/js/src/mixins/SearchUtils.js

Code scanning
Incomplete string escaping or encoding
This replaces only the first occurrence of '\n'.

If i make 2 line breaks and do bad stuff on the last line, its not caught.

So this: 

mus fisk AND
test AND test
and test

Is not caught at the moment, at least not when I tried it.

jesperlauridsen (commented):
I've only just found the one bug, that codescanner also found. Just check if that checks out or not, and i'll approve if its not a bug. Other than that, this looks freaking awesome, well done by both of you!

src/js/src/mixins/SearchUtils.js
tokee: Nice catch! If should be `q = q.replaceAll("\n", "*");` (or we could switch to regexp with `q = q.replace(/\n/g, "*");` as the `/g`-modifier only works with regexp, but since this is simple single-char replacement, there is no need for the switch).

jorntx: Thank you Code scanner! I agree replaceAll would be the best and simplest path here. But why switch to replacing with "*" it was ' ' before @tokee?

jesperlauridsen (approved):

src/js/src/mixins/SearchUtils.js
tokee: Dóh! That was just me testing in a way where I could see the result. It should of course be ` `. Sorry.

PR #157, opened 2020-11-27
fix to encoding error.

jesperlauridsen:
removed encoding when doing a new search (from that specific method) since it's now taken care of when the search is performed (in the requestService.js file).

@thomasegense  would you verify that this takes care of the problem with the document fields in the foldout.


thomasegense (approved):
Seems to work. But when more branches with fixes are merged into master, a new complete monkey test is required

PR #160, opened 2020-12-02
Soback 98 better load display for images loading

jesperlauridsen:
Images shown in image search now displays a turning wheel when the image is loading

Images shown in a post now displays a turning wheel when the image is loading

codingwise, @jorntx, i've just moved a little bit of stuff around to be able to have a new div in each of the templates containing an image. Nothing too fancy - but took some work since the focused full image is handled by the parent in both scenarios.

@thomasegense for functionality.

### DIFF
--- a/src/js/src/assets/styles/results.scss
+++ b/src/js/src/assets/styles/results.scss
@@ -506,14 +506,24 @@ margin-left:40px;
 }
 
 .singleEntryImages .previewImageContainer {
+  position: relative;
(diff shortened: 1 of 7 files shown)


jorntx (approved):
Works beautifully 👍 

I haven't found any issues with the code - as you wrote in your comments most of it is reused/moved code.

thomasegense (approved):
OK

PR #162, opened 2020-12-04
Quote regex replaced with version without look behind due to Safari issues

jorntx:
@tokee and @thomasegense you can just check that it  (still) does what you want.

I'm the only one with Safari at the moment and I can confirm that the app now starts.


thomasegense (approved):
Still working for me

tokee (approved):
Works as intended. Go ahead!

PR #163, opened 2020-12-04
Fixed error where encoding was done when it shouldn't be.

jesperlauridsen:
This was wrongfully added again - possibly because of a mixup with what branches was merged into master first and last - Removing this encoding *should* make all well again.


thomasegense (approved):
Looks fixed,also tested other queries still working.

jorntx (approved):
Works fine but haven't I seen this exact code in this exact place before? Degeneration? 🤔

PR #178, opened 2021-06-07
Better index-script for Linux/OS-X

tokee:
This adds a bash-script that

 * Supports easy parallel processing (default is 2 processes in order to avoid blowing the memory)
 * Stores a log file for each processed WARC
 * Cleans up temporary files automatically
 * Keeps track of which WARCs are indexed
 * Has rudimentary check for failed indexing (a lot of room for improvement here)

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ Unzip and follow the instructions below.
 
 ### 1) INITIAL SETUP  
 Properties:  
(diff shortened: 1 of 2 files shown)


thomasegense:
Make  tmp+log files configurable. 

Warc-file collection (preservation) should not be mixed with temp files. Also problem with write access to the archive.

tokee:
> Warc-file collection (preservation) should not be mixed with temp files. Also problem with write access to the archive.

Very true. Latest commit fixes this and adds a check to see if Solr is running before attempting to index anything. The fix does require WARC file names to be unique, but it does check for uniqueness upon startup and complains if duplicates are found.

PR #183, opened 2021-08-22
Modify < > handling to preserve the original representation

tokee:
Handling `<` (and to a lesser extend `>`) in inline JavaScript in XHTML is messy as `<` is not allowed directly inside of `script`-blocks. This was at some point handled by wrapping as `CDATA`, but with HTML `script`-block content is always treated as `CDATA` so the problem went away. The old code tried to handle all this by always escaping to Unicode (`\u003C`), but this produced illegal JavaScript.

This fix tries to preserve the original representations instead. There might still be errors as JSoup is in the mix and does some unescaping while parsing the content but the current unit tests pass.

This closes Royal Danish Library internal issue SOLRWBFB-58


thomasegense (approved):
It fixed the console log error on 20+ different sites I checked.  

Nice with unittests :)

PR #190, opened 2021-09-10
fixes #189 - load more facets bug and removed unwanted console.log

### LINKED ISSUE #189: Load more facets fail to load if any facets already are applied
Load more facets fail to load if any facets already are applied.

Steps to reproduce:

1. Make a search
2. Select a facet
3. Try to load more facets in any relevant facet category

jorntx:
Should fix the load more facets bug.

Fixed by splitting the encoding of the main query from the facets.

Also removed an unwanted console.log (route chenged!)


thomasegense:
looking goood

Issue #88, opened 2020-09-17
Twitter json-warcs,find hashtags from json indicies.


thomasegense:
The code finds hashtags now by searching for words that starts with hashtag. This is not always correct, instead
use the incides for hashtags start-end in the JSON.

This Jira has very low priority. (twitter-warcs is custom format not used by others)

Example json:
"hashtags":[
               {
                  "indices":[
                     273,
                     279                 
],
                  "text":"math"               
}

thomasegense:
@Bohlski  fixed this. Closing.

PR #199, opened 2021-11-09
More Twitter playback

Bohlski:
Added solution for #88, better testing and some refactoring


thomasegense:
I reviewed most of the stuff before this PR. I will accept the PR and do some testing with latest on master.

PR #202, opened 2021-12-10
Corrects MIME type when supplied type is application/video

jorntx:
Fixed the MIME type in the cases when Tika supplies the non standard application/video


thomasegense (approved):
Simple fix :)

thomasegense:
Confirmed fixed.

PR #210, opened 2022-03-04
WARCs over HTTP

tokee:
@anjackson needs support for WARCs over HTTP and some digging around for a HTTP InputStream implementation with efficient skipping (using [HTTP Range Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests)) did not give any results.

This pull request adds an implementation of efficient HTTP skip; the `ArcHTTPResolver`. Skips are handled by closing the existing HTTP connection and opening a new one at `<current position> + <skip>`.

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/interfaces/ArcFileLocationResolverInterface.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/interfaces/ArcFileLocationResolverInterface.java
@@ -1,5 +1,6 @@
 package dk.kb.netarchivesuite.solrwayback.interfaces;
 
+import java.util.Map;
 import java.util.function.Supplier;
 
 public interface ArcFileLocationResolverInterface {
@@ -22,7 +23,7 @@ public interface ArcFileLocationResolverInterface {
    */
   ArcSource resolveArcFileLocation(String source_file_path);
 
-  void setParameters(String parameter);
+  void setParameters(Map<String, String> parameters);
   void initialize();
 
 }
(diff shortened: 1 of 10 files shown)


anjackson:
Thanks, but unfortunately that currently conflicts if I try to merge into the `solr_params` branch. @thomasegense can we bring `solr_params` into `master` or is there more work that needs doing first?

thomasegense:
Hi Toke. I get a unittest error:
Tests in error: 
  testSkipper(dk.kb.netarchivesuite.solrwayback.util.SkippingHTTPInputStreamTest): Unable to (re)connect and skip 31744 bytes from 32768 to 64512. Only skipped 0 bytes to 32768: https://labs.statsbiblioteket.dk/anskuelse/adams_history/18068_Adams_Illustrated_Panorama_of_History_1878_size_104345x11288.tif
  testSkip(dk.kb.netarchivesuite.solrwayback.interfaces.ArcHTTPResolverTest): Unable to (re)connect and skip 31744 bytes from 32768 to 64512. Only skipped 0 bytes to 32768: https://labs.statsbiblioteket.dk/anskuelse/adams_history/18068_Adams_Illustrated_Panorama_of_History_1878_size_104345x11288.tif

thomasegense:
@anjackson I will merge into master later today if everything goes well.
I am making the url_norm to url switch. We found an easy solution, but it has not been testet it will work.
Using the new solr-params you set the additional property:
f.url_norm.qf=url

And I am making a property to disable normalization. So that is just one place in the code I have to fix and use the property

PR #212, opened 2022-03-07
Solr params

thomasegense:
This branch has code to disable normalisation and use url instead of url_norm for playback.
This can be improved further by making a heritrix normaliser. since space +, %20 in webpages are normalized by heritrix in the warc-header.


thomasegense:
Closing the branch now. 
Normalisation class is now defined in solrwayback.properties

PR #215, opened 2022-03-09
Add test WARCs, with indexer and server

anjackson:
This attempts to:

- Provide test WARCs rather than just Solr records.
- Populate the index directly using `warc-indexer` when populating the index.
- Includes a `warc-server` so the `ArcHTTPResolver` back-end can be tested.
- Configures the service to use `ArcHTTPResolver` and `url` instead of `url_norm`.

It does not fully work at present, as some critical queries appear not to add the`solr.search.params` because it tries this (0 hits):

http://dev1.n45.wa.bl.uk:18983/solr/discovery/select?indent=on&q=url_norm:%22http://acid.matkelly.com/%22+AND+status_code:200&defType=edismax&qf=text&fl=id,score,title,url,url_norm,links_images,source_file_path,source_file,source_file_offset,domain,resourcename,content_type,content_type_full,content_type_norm,hash,type,crawl_date,content_encoding,exif_location,status_code,last_modified,redirect_to_norm&pf=text&fq=record_type:response+OR+record_type:arc+OR+record_type:resource&wt=json 

But this works (1 hit):

http://dev1.n45.wa.bl.uk:18983/solr/discovery/select?indent=on&q=url_norm:%22http://acid.matkelly.com/%22+AND+status_code:200&defType=edismax&qf=text&fl=id,score,title,url,url_norm,links_images,source_file_path,source_file,source_file_offset,domain,resourcename,content_type,content_type_full,content_type_norm,hash,type,crawl_date,content_encoding,exif_location,status_code,last_modified,redirect_to_norm&pf=text&fq=record_type:response+OR+record_type:arc+OR+record_type:resource&f.url_norm.qf=url&wt=json


anjackson:
Ah, okay, that issue appears to be with having multiple Solr params. In short, this:

```
solr.search.params=f.url_norm.qf=url;defType=edismax;pf=text;qf=text
```

does pick up the `url_norm` mapping, but this (same args different order) does not:

```
solr.search.params=defType=edismax;pf=text;qf=text;f.url_norm.qf=url
```

anjackson:
Ah, also, the proposed ArcHTTPResolver regex wasn't working as expected.  This works for me:

```
warc.file.resolver.parameters.path.regexp=.*/([^/]*)$
```

whereas in the supplied form:

```
warc.file.resolver.parameters.path.regexp=.*([^/]*)$
```

the `.*` at the start is greedy and consumes the whole string.

anjackson:
This committed version now works. There appear to be some playback bugs, but image search seems to work, for example.

PR #219, opened 2022-03-25
Updates from running at UKWA

anjackson:
This covers a few different issues I hit when getting SolrWayback running at UKWA as an internal service.  If you'd rather they were split up into separate PRs, let me know.

It includes:

- A GitHub Action to build a Docker image
- The ability to map in whole properties files for configuration rather than only configuring a few bits via env vars.
- The ability to configure a different alternative playback URL when running via the Docker image.
- An alternative filename regexp that copes with or without a leading path.
- The abililty to run under a sub-path, like `/prefix/solrwayback/`.

That last one might be problematic.  Making the WAR deploy at a prefix was easy enough as there is a `prefix#solrwayback.war` naming convention for this.

The hard part was making Vue happy. Reading around, I found I could make this work, based on this: https://stackoverflow.com/a/69275528/6689 . However, the documentation indicates that this feature may not work with `history.pushState`. See the 'limitations' section of https://cli.vuejs.org/config/#publicpath

I'm not 100% sure that _all_ features are still working, and I'm not sure I'm aware of them all.  And I'm not sure how much of a problem the above limitations are. But apart from that it's all great! :-)

### DIFF
--- a/.github/workflows/push-to-docker-hub.yml
+++ b/.github/workflows/push-to-docker-hub.yml
@@ -6,6 +6,7 @@ on:
       - '*'
     branches:
       - master
+      - docker-hub-action
 
 jobs:
   push_to_docker_hub:
--- a/.github/workflows/push-to-docker-hub.yml
+++ b/.github/workflows/push-to-docker-hub.yml
@@ -1,0 +1,34 @@
+name: Run CI then push to Docker Hub
+
+on:
+  push:
+    tags:
+      - '*'
+    branches:
+      - master
+
+jobs:
+  push_to_docker_hub:
+    name: Push Docker image to Docker Hub
+    # This uses the default metadata-action configuration, see https://github.com/docker/metadata-action
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Set up Docker metadata
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: ${{ github.repository }}
+      - name: Log in to Docker Hub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ Docker support. (so far limited support... write more later)
 Support for legacy WARC-Indexer before version 3.0 that does not have url_norm field. See below. 
 New property in solwayback.properties 'url.normaliser'. Will default to normal. Other options are minimal and legacy. 
 New property in solwayback.properties 'solr.search.params'. Add default solr params to every query.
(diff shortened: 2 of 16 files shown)


thomasegense:
Hi AndyThis will take a little time to test since it can break some functionality.
But the alternative playback was already implemented but setting the property:
## Optional additional playback engine. 
openwayback.baseurl=http://web.archive.org/web/
Of course the name "openwayback' is misleading. The alternative playback gives an icon next to the title. 
It is because you want the primary playback URL to point somewhere else than to solrwayback playback?

anjackson:
Sorry, I didn't write that write - you already had the support for alternative playback, I just made it easy to override when deploying via Docker!

thomasegense:
I have tested most corners  in SolrWayback  with our setup/property-files and no context root change. And it does not break anything :) So you fixed it so it worked with a context root change that is excellent.   I will merge the branch and before next
bundle release I will do a complete UI test again.  

Docker does still not images/binaries and playback. 
Seems there still is an regexp error.

 ERROR StatusLogger Unrecognized format specifier [d]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [d] starting at position 16 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [thread]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [thread] starting at position 25 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [level]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [level] starting at position 35 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [logger]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [logger] starting at position 47 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [msg]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [msg] starting at position 54 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [n]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [n] starting at position 56 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [d]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [d] starting at position 16 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [thread]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [thread] starting at position 25 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [level]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [level] starting at position 35 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [logger]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [logger] starting at position 47 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [msg]
populate_1     | ERROR StatusLogger Unrecognized conversion specifier [msg] starting at position 54 in conversion pattern.
populate_1     | ERROR StatusLogger Unrecognized format specifier [n]

jorntx:
> Sorry, I didn't write that write - you already had the support for alternative playback, I just made it easy to override when deploying via Docker!
> 
> _EDIT_ I've done a tiny bit of Vue but really I'm just hacking stuff in, so I'm glad someone else will look at it. If this doesn't work, I think we can switch to a model where we build a container image per deployment. It'll be a bit clumsy bit I think it should work.

The Vue part seems to work fine and passes my quick sanity check. Nice fix with `location.pathname` in the router!

anjackson:
Hah thanks @jorntx but it's really all down to my _Cut & Paste from Stack Overflow_ skillz!

PR #222, opened 2022-04-08
Fix ngram end year bug

jorntx:
Fixes ngram end year bug by making the end year the current year. For some yet unknown reason we have to +1 the year to make the graph tool work.


thomasegense:
Looking gooooood!

PR #225, opened 2022-06-09
Fix for Twitter playback to show indexed videos as part of tweet.

Bohlski:
Also reworked whole Twitter JSON-parsing, so it is now done using FasterXML's Jackson. 

Making this PR now, I realize that Gson might've been a better alternative for JSON-parsing (just read this: https://www.overops.com/blog/the-ultimate-json-library-json-simple-vs-gson-vs-jackson-vs-json/), but as I was only familiar with Jackson when starting this rework, I simply went with that. 
Imo Jackson is still miles better and more readable than what I had done before using the lightweight JSON-Java package.


thomasegense:
Accepted. Overall seems good.

PR #238, opened 2022-09-05
GitHub 232 search does not handle params well

jorntx:
Encoded URL param in 'normalize url' request


thomasegense (commented):
simple

thomasegense:
Accepted. This PR is required for the next bug fix involving correct encoding of the url_norm query fired after the normalize service call

PR #239, opened 2022-09-05
Make modals and image views close on escape event

jorntx:
First stab at making modals and expanded image views close when hitting the escape button


thomasegense:
Working  for toolbox, error message, images. Can  not find anything missed.

thomasegense:
closed

PR #240, opened 2022-09-05
Fix inconsistent testing by explicit reset of URL normaliser

### LINKED ISSUE #237: Test problems under some platforms
The current version of SolrWayback (master branch, 2022-09-02) fails under Ubuntu 22.04.1, OpenJDK 11.0.16 with
```
Failed tests:   testScriptRewriting(dk.kb.netarchivesuite.solrwayback.parsers.HtmlParserUrlRewriterTest): The result should be as expected for test 'script' expected:<...fdgfnijoic3zlx3nkin0[\]u00253d&_nc_ohc=pnym...> but was:<...fdgfnijoic3zlx3nkin0[%5c]u00253d&_nc_ohc=pnym...>
  testExternalScript(dk.kb.netarchivesuite.solrwayback.parsers.ScriptRewriterTest): expected:<...fdgfnijoic3zlx3nkin0[\]u00253d&_nc_ohc=pnym...> but was:<...fdgfnijoic3zlx3nkin0[%5c]u00253d&_nc_ohc=pnym...>

Tests run: 71, Failures: 2, Errors: 0, Skipped: 0
```

It does not fail under Ubuntu 18.04.6, OpenJDK 11.0.11.

Quick guess is that some part of the code is dependent on the local environment for choosing String encoding upon data retrieval.

tokee:
Some unit tests are reliant on the configuration being in a certain state. The configuration is static, so the order of unit tests mattered. As the order of execution of unit test classes is not fixed, this caused testing to fail on some setups.

This pull request adds an explicit reset of the configuration to tests needing consistent normalisation of URLs. It won't work for parallel execution of unit tests (which we don't use at this point in time).

This closes #237


thomasegense:
tricky one. 
Merged... Another issue to validate if the different behavior between legacy and normal is expected. Requires checking warc-indexer.

PR #242, opened 2022-09-06
Refactor normalization

thomasegense:
This PR fixes some bugs with normalisation. Both in unittest but also in the application.

The bug fixes normalisation in case normaliser is legacy and the domain is a domain without part.
Legacy normalizer
"http://www.test.dk/" -> "http://www.test.dk/"  (this was the bug)
"http://www.test.dk/index.html" -> "http://test.dk/index.html"  (here  www is removed.)

The bug was also fixed in the url-search method that both punyencodes and normalise the url


thomasegense:
And merged.

PR #243, opened 2022-09-09
Quote phrase URL searches and handle data: URLs

### LINKED ISSUE #230: Website does not show correctly
Hi,

I have a problem with some WARC files in my collection. For example, [this file](https://drive.google.com/file/d/18HlPVOWFf_QM4cY3BkI3cP4A_7Fu4rng/view?usp=sharing) I have from [www.pvda.nl](https://www.pvda.nl) does not display correctly. I checked whether the problem is in the file itself, but it seems to be okay. When I load it in pywb, it looks like this:
pvda pywb

Whereas when I load it in Solrwayback, it looks like this:
pvda solrwayback

Somehow, it looks like it is not loading the CSS and image files. When I look in my browser's console, I see a bunch of 404 errors. I tried it in Safari, Chrome and Firefox, but that does not make any difference.

The file can be found here: https://drive.google.com/file/d/18HlPVOWFf_QM4cY3BkI3cP4A_7Fu4rng/view?usp=sharing

Can you help me solve this problem? Thank you.

tokee:
This closes #230 . Sorry about the extremely messy diff: I did it to get the pull request working as it did not like `master` having the same changes reverted. I need to get better at `git`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,8 @@
 4.3.1
 -----
 Playback improvement. Fixed some redirect/url-parsing bugs in the ROOT.war (solrwaybackrootproxy). 
-TODO: Fix the solr-escaping+ data-image bug...
+Playback improvement: Queries for page resource resolving are now properly quited, avoiding a scenario where resolving of all page resources failed. https://github.com/netarchivesuite/solrwayback/issues/230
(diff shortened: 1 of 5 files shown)


thomasegense (approved):
Good simple fix. The webpage in questions now playbacks perfect.

PR #249, opened 2022-09-12
Support for data-src

### LINKED ISSUE #247: Playback url-replace for 'data-src' property (minor improvement)
Some image-tag has data-src property and this can be url-replaced up front when serving the HTML page as the other url-replacements.
However playback does work since the serviceworker will catch it and feed it back into SolrWayback and resolve it properly.
The fix will mean service does not have to do any work.

Below is an example with 'data-src'

The warc-file listed in #230 has this on same webpage as mentioned.

``

tokee:
JQuery conventionally uses [data-src for delayed image loading](https://learn.jquery.com/code-organization/deferreds/examples/) (``). This pull request adds support for playback of webpages using that convention.

This closes #247


thomasegense:
Confirmed that data-src has been url-replaced

PR #252, opened 2022-09-13
Cache invalidation

### LINKED ISSUE #198: Light weight cache invalidation
An obvious pattern when using the SolrWayback bundle is to check indexing progress using SolrWayback and the internal SolrWayback caching of results means outdated responses. the cache should be invalidated when the number of indexed documents changes.

Unfortunately even a "null"-query such as `q=*:*&rows=0&facet=false&group=false&hl=false` is non-trivial for billions of documents. It seems reasonable that there should be _some_ way of detecting if the index has been changed, but how!?

One solution would be to have a heartbeat checker running in the background, invalidating all caches if the number of documents in the index changes. But this risks outdated results and puts a recurring load on the index.

tokee:
Add timer based cache invalidation by checks for index change, as suggested in #198 

Also adds webservice endpoint for checking Solr availability - maybe this could be used from the SolrWayback application in some way?

This closes #198

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,7 @@ issues.
 GUI improvement. Use escape-key to close all modal pop-ups. (toolbox, search syntax, full size 
 images)
 
+Cache invalidation based on background checks for index changes.
 
 
 4.3.0
--- a/docker/solrwayback.properties
+++ b/docker/solrwayback.properties
@@ -6,7 +6,7 @@ solr.server=${SOLR_URL}
 #Solr caching. Will be default false if not defined
 solr.server.caching=true
 solr.server.caching.max.entries=10000
-# Max age is set to a year, effectively disabling it, as solr.server.check.interval is used.
+# Max age is set to a year, effectively disabling it, as solr.server.check.interval.seconds is used.
 solr.server.caching.age.seconds=36586400
 
 # Solr availability and index change check interval.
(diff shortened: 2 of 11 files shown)


tokee:
I sanity checked performance against our own archive:
```
curl 'http://localhost:52300/solr/ns/select?q=*:*&rows=0&stats.field=\{!max=true\}index_time&stats=true'
```
took 102 seconds, while querying without changes with
```
curl 'http://localhost:52300/solr/ns/select?q=index_time:\{2022-08-30T04:11:06.626Z%20TO%20*\]&rows=1&fl=index_time&facet=false'
```
took ~100ms (Solr query cache so no real load on Solr) and simulated querying with changes
```
curl 'http://localhost:52300/solr/ns/select?q=index_time:\{earlier_timeZ%20TO%20*\]&rows=1&fl=index_time&facet=false'
```
took 15-30 seconds, depending on the _earlier_time_ value.

With background startup of the `IndexWatcher` and the _frozen shards_ setup at the Royal Danish Library this works perfectly well. For a "large" archive (billions of records) with a _hot shards_ strategy and continuous updates, using the IndexWatcher might be too expensive. I will update the sample configurations accordingly with a note on this.

thomasegense (changes requested):
The properties for solr-caching are seconds and not millis.
Change the property to use seconds as well:
solr.server.check.interval=60000

Can you remove the obsolete property: solr.server.caching.age.seconds=36586400 ?
(and remove logic+loading in PropertiesLoader)

thomasegense:
Tested and merged

PR #253, opened 2022-09-13
Remove data: URLs from the raw list of page resources

### LINKED ISSUE #250: Remove data: links from Missing Resources page
When listing "Missing Resources" under tools, `data:`-links shows up. As these are HTML-enbedded resources, they are not missing and should not be listed.

tokee:
Small fix, no unit test (sorry).

Maybe the skipping of `data:` URLs should be moved down to the `HtmlParserUrlRewriter.getResourceLinksForHtmlFromArc` instead? Would that break any other functionality?

This closes #250


thomasegense:
Confirmed the data: was removed. 
But this fix seemed special casing and not using the tags/attributes method that. Was it not possible?

PR #265, opened 2022-10-05
Add support for access control list

krakan:
We (National Library of Sweden) would like to have an access control list to block access to certain URL:s. Supposedly this is of interest to other organizations as well.
It's very likely that this can be improved, feel free to poke holes in my code.

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/properties/PropertiesLoader.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/properties/PropertiesLoader.java
@@ -48,6 +48,8 @@ public class PropertiesLoader {
     
     public static final String PLAYBACK_DISABLED_PROPERTY="playback.disabled";
     private static final String SOLR_SEARCH_PARAMS_PROPERTY="solr.search.params";
+    private static final String ACCESS_CONTROL_LIST_PATH_PROPERTY="access.control.list.path";
+    private static final String ACCESS_CONTROL_HEADER_NAME_PROPERTY="access.control.header.name";
     
     
     private static Properties serviceProperties = null;
@@ -63,6 +65,8 @@ public class PropertiesLoader {
     public static String PID_COLLECTION_NAME = null;
     public static String WORDCLOUD_STOPWORDS;
     public static LinkedHashMap<String,String> SOLR_PARAMS_MAP= new LinkedHashMap<String,String>(); 
+    public static String ACCESS_CONTROL_LIST_PATH = null;
+    public static String ACCESS_CONTROL_HEADER_NAME = null;
 
     public static boolean SOLR_SERVER_CACHING=false;
     public static boolean WARC_FILES_VERIFY_COLLECTION=false;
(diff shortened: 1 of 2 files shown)


krakan:
There most probably need to be a few test cases and additions to the docs. I can fix that in that case.

thomasegense:
Hi. I am on vacation the next week but will try answer without my computer.

There already is a way to block stuff like url's or domains in
Solr.(schema.xml).
@Toke Eskildsen ***@***.***>  can maybe show you the snippet.
Blocking it once in Solr will prevent and search result+playback for those
urls.

krakan:
Yes, I had seen that you could put filters into the Solr config but I found that hard to combine with a role-based approach.
I'll look into using solr-filters instead of my current approach of matching the response against a bunch of strings.
And, yes, making sure there are no ways to circumvent the ACL will take quite a bit of work.

tokee:
> ... making sure there are no ways to circumvent the ACL will take quite a bit of work.

First of all, I might have misunderstood what you are trying to do. Is is only _playback_ that you want to control? Is so, then your approach seems solid enough. If you also need to avoid indirect information leaking, such as image search or export, then we will have to perform the thorough review.

Guessing you need the full block, I'll just note two ideas here:
1. All (at least that's hos it should be) Solr requests using `SolrQuery` passes through `NetarchiveSolrClient.setSolrParams` before being used. With the filtering approach it should be possible to hook into the request chain just there instead of "everywhere". Be aware that some requests might not call this method, either due to not using `SolrQuery` or due to an oversight. 
2. There is already a `CachingSolrClient` that is an extension of `SolrClient` with client side caching. SolrWayback **always** uses a `SolrClient` when communicating with Solr, so it might be less error prone to hook the ACL into the `SolrClient`s that are used by creating a wrapper that either adjusting input (by ading a filter) or output (by pruning documents). Do note that adjusting output would require supporting at least plain and grouped searches.

krakan:
Ah, well, yes, the patch is indeed only intended to block playback. But as you point out this means that you can gain information about blocked data simply by searching for it. My first approach was to ignore that issue for now.
After having talked about this with colleagues, we're leaning towards not yet being allowed to allow full text search at all, as the laws governing our collection aren't totally clear and are - I've been told - in the process of being reformulated.
This means we have no need for any fine grained ACL until the legal ramifications have been clarified.
The plan is still to eventually use SolrWayback and your ideas do seem interesting so I'll try to find time to implement them anyways.

krakan:
Hm, tricky ... What one would have to do is to pass the HttpServletRequest (or just the authenticated user name from within the request) as a parameter through all chains of method calls from the SolrWaybackResources down to the SolrClients so that the client can choose which set of filters to apply. Doable but feels kind of messy.

An alternative approach would be to have multiple SolrWayback-instances, listening on different ports but talking to the same Solr-server, and then let the reverse proxy server (which is already responsible for authenticating the users) choose which instance to talk to. That would consume more resources but would be somewhat less messy as long as the number of instances is small. In our case we would get away with just two.

Being able to update an ACL-list without restarting the SolrWayback server is an independent problem that is easier to solve.

tokee:
We could use [ThreadLocal](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html) to hold the `HttpServletRequest` and extend the `Solrclient` to use that for access validation, but that would require another contract: That the request for a `SolrClient` happens in the same Thread that was created when the endpoint was called. It seems fairly easy to implement as we wouldn't have to send the `HttpServletRequest` through all methods, but also more prone to errors when things like threaded image lookups are being activated.

\<Thinking pause\>

If we add a [Servlet Filter](https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/Filter.html) to SolrWayback that stores the `ServletRequest` in `ThreadLocal`, that should ensure it is always set for the initial Thread. Should some code mess it up and request a `SolrClient` from a Thread that is not the initial one, it will get `null` and fail.

tokee:
Okay, people don't seem to like the `ThreadLocal` idea: https://stackoverflow.com/questions/5473077/threadlocal-to-store-servletrequest-and-response-in-servlet-what-for. And it's a fair point that it is messy to mix container code with regular code.

thomasegense:
This solution is specific for you need, so I will look into a more general solution. So far there only is a switch to activate playback for all users.  Solr has an effective switch to block all content that can be expressed as a query, but again for all users.

The tricky part is decide playback quests from search queries. When there are more requests from the user base, I will try see what the more general need it. But I would like to hear if you had success with solution.

PR #267, opened 2022-10-14
Maintenance 20221014

jorntx:
Updates npm packages and bumps node to 6.18.0 and npm to 8.19.2.

Major version upgrade of videojs.

Focus on function testing and check if Maven build is still OK.


thomasegense (approved):
Could not find any broken stuff. Everything seems to work.

Issue #270, opened 2022-10-19
Refine time frame and time scale on "Visualization of search result by domain"


saraaubry:
Currently, the "Visualization of search result by domain" graph is distributed by year (1998-2022). The display is not useful when using SorlWayback with a smaller collection. In our case, the collection ranges from march to june 2022.

BnF suggests an evolution to give users the possiblity to refine :
- the time frame using 2 dates defined as YYYY-MM-DD -> YYYY-MM-DD.
- the time scale (with a choice between year, month, week, day)

Defaults value are : 
- date 1 : defined setting for the beginning of the archive (1998-01-01)
- date 2 : today
- time scale : year
which do not change current implementation.

![visu-search-results](https://user-images.githubusercontent.com/6735961/196646874-73e95e88-0249-4c97-89ce-ff7c4288e9b9.PNG)

thomasegense:
That is a very good idea for the GUI! 
Right now you can only configure the start year in the solrwaybackweb.properties. The default is 1998. 
But for newer collections using months as buckets makes a lot more sense and you are not the first to suggest this.

I will make this issue high priority and hope we can implement it soon.

The minor details is we do not have year-month is not in the index. The current implementation uses crawl_year. And it took two optimizations each giving a factor 10, to make it perform as it is now.  So performance may be a lot worse when using months as buckets. But for large and old collections you probably do not want to use months anyway. And newer collections are much smaller so performance might not be an issue. The date field crawl_date can maybe be used for this.

I hope @tokee  can find a way to make Solr give this dataformat. I will do the implementation.

bnfleb:
I already implement the Solr query using the field crawl_date:
`q=*:*&rows=0&facet=false
&json.facet={
    domains:{
        type:terms,
        field:domain,
        limit:30,
        facet:{
            years:{
                type:range,
                field:crawl_date,
                start:'2002-01-01T00:00:00Z',
                end:'2003-01-01T00:00:00Z',
                gap:'+1MONTH'
                }
            }
        }
    }`

I need a little time for the GUI but the PR will come soon

thomasegense:
@tokee  I will try the query tomorrow. Starting with a smaller year range. But I think you could be right that some validation of input can be required. Maybe depending on total number docs?

thomasegense:
@bnfleb  @tokee  I did some performance testing with the gap size DAY or MONTH. The test was against our full corpus (40B documents)

[...]

PR #272, opened 2022-10-21
Refine time frame and time scale on "Visualization of search result by domain"

bnfleb:
As discussed in issue #270, I have added the possibility of refining the time frame and time scale on "Visualization of search result by domain".

The first graph is the same as before (using crawl_year) but by refining the field crawl_date will be used.

For now, there is no check (except for the date format) because it seems like there is no blocking checks in the features that are in a modal window. 
For the time scale, an automatic selection according to the time frame can be added to prevent a too big range for a gap size DAY or MONTH

### DIFF
--- a/src/js/src/assets/styles/results.scss
+++ b/src/js/src/assets/styles/results.scss
@@ -297,7 +297,8 @@ margin: 1em 0 0 0;
 .exportContent .downloadButton, 
 .wordcloudButton, 
 .domainContentContainer .domainStatsButton,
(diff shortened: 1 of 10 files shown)


thomasegense:
Thanks, I will test it next week and hope @jorntx  and myself will have some time to look into it.
Will be interesting.

jorntx (approved):
Solid frontend work - approved!

thomasegense:
And backend code also approved! No further changes required.

PR #276, opened 2022-10-28
GitHub 271 export encoding error

jorntx:
Adds encoding to all search result exports.

Changed 'URL building' from string concatenation to template literals for all search result exports


thomasegense:
Fixed both know encoding bugs. ( [] and +).
Thanks

Issue #255, opened 2022-09-14
Optional lenient URL resolver


tokee:
Heritrix has problems with `<img srcSet...>` support: https://github.com/internetarchive/heritrix3/pull/179/ resulting in attempted harvesting of `img1%20img2` instead of `img1` and `img2`. Surprisingly it works with some image servers, which will return the image data for `img1` when `img1%20img2` is requested (Guessing: Because of the space).

`NetarchiveSolrClient.findNearestDocuments` is used for locating page resources (images, CSS etc.) for playback. We should consider adding an option for lenient resolving: Start by searching for the URL directly and if that fails, split on first space (maybe also comma and semicolon?) and do a wildcard search on the first part, e.g. `img1*`.

tokee:
This seems related to URLs that ends with `&_=123456789` (used for invalidating browser cache), which can also present a problem. Maybe the lenience handling could be rule based and suggest alternative lookups depending on the originating URL?

```
"([^ ]*) .*" → "$1*"
"(.*)&_=[0-9]+" → "$1*"
```

tokee:
Thinking more about this: The field `url_search` tokenized the URL and should be usable for this.

Let's say we have the URL `http://example.com/images/search?q=horse&fq=animals&_=12345` indexed and we are looking for `http://example.com/images/search?q=horse&fq=animals&_=67890` (the ending is different).

We consider the domain and the path to be fixed, with all arguments as optional. This query is possible with
```
( domain:"example.com" AND url_search:"/images/search" )
OR url_search:"q=horse"
OR url_search:"fq=animals"
OR url_search:"_=12345"
```

Underneath the hood this will be
```
( domain:"example.com" AND url_search:"images search" )
OR url_search:"q horse"
OR url_search:"fq animals"
OR url_search:"12345"
```
so the match is not ideal, compared to dedicated argument indexing, but it matches pretty well and (important here) is quite fast and with usable ranking.

PR #281, opened 2022-11-02
Lenient url search

tokee:
Adds `UrlUtils.lenientURLQuery(url)` which takes an URL and creates a query that accepts partial match on arguments in the URL, as described in #255. The code is not tied to anything at the moment, as it is not clear when we want to be lenient.


tokee:
The lenient URL search is enabled for playback using the (currently secret) endpoint `lenient/web/` instead of just `web/`.

tokee:
Fixed the bug with missing resolving of URLs when there were > 1000 instances of the same URL and added optimization of URL lookups using Solr grouping when possible.

tokee:
Solr grouping are now used all the time when deduplication is enabled. I have nothing further to add in this pull request.

PR #288, opened 2022-11-15
Refine time frame and time scale on domain stats

bnfleb:
Same principle as for #272, now the user have the possibility to refine the time frame and time scale for the domain statistics.

I added a check : if the time frame contains more than 90 buckets (years, months, weeks, days), the graph is not generated (no solr interrogation) and a message is displayed.
The limit is fixed but can be added as a parameter.

### DIFF
--- a/src/js/src/assets/styles/results.scss
+++ b/src/js/src/assets/styles/results.scss
@@ -1185,6 +1185,10 @@ margin-left:40px;
   margin: 0 5px;
 }
 
+.timePeriodRefinerTimeScaleContainer select {
+  margin-left: 0;
+}
+
 .modalContainer .closeButton, .toolboxCloseButton {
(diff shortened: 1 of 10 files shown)


thomasegense (commented):
Backend code approved.
It was correct to move the DomainStatistics class to the other DTO's. Nice find.

bnfleb:
One more thing, search with a different time scale returns a different total number of pages because Solr calculate the number of unique pages for the time scale. A page archived in two different months will be counted as 1 page for a year time scale but as 2 pages for a month time scale.

It may be usefull to add somewhere that the result displayed is the number of **unique** pages for the period searched.

jorntx (approved):
Follows along the lines from the previous similar PR - still looks.

PR #326, opened 2023-03-23
Changes export url handling

jorntx:
Changes export url building from string literals to string concatenation. Fix bug when exporting to WARC and CSV.


thomasegense:
merged

PR #328, opened 2023-03-24
Add encoding to filters when exporting

jorntx:
Added encoding to filters when exporting


thomasegense (approved):
Tested and still working.

Issue #218, opened 2022-03-21
Add support for stats


tokee:
The [Solr stats component](https://solr.apache.org/guide/8_11/the-stats-component.html) delivers statistics, primarily for numeric fields, such as average, count and sum. It is benign in its default state, meaning that it does not have excessive memory overhead. Extra features such as `cardinality` are a bit dangerous and the `distinctValues` & `countDistinct` will result in Out Of Memory for some fields.

SolrWayback should add support for calling the stats component in a controlled manner, meaning that only fields & features specified in the setup should be allowed.

VictorHarbo:
I would like to take a look at implementing this.

VictorHarbo:
#330 implements the Solr Stats component in the java layer. To fully implement this functionality these functions have to be called and used by the frontend. 

Some following steps:

- Create java facade for `SolrStats`.
- Figure out where to call this in the frontend

VictorHarbo:
With #339 merged this issue is down to one unresolved feature: How to implement the stats component in the frontend. 

Stats queries can have a heavy load time, which needs to be communicated to the end-users in a meaningful way.

jesperlauridsen:
If we already have a open issue for this #343, can we safely close this one, @VictorHarbo?

VictorHarbo:
Yes! Indeed we can! If you feel like it and have the time, you are welcome to have a look at #343.

VictorHarbo:
Closing the issue as the backend part of the stats component has been implemented and the frontend has its own issue: #343

PR #330, opened 2023-04-17
218 implement stats

VictorHarbo:
First check-in for implementing the solr stats component as requested in #218.
This PR implements three methods in the java class SolrStats: 

- getStatsForMultipleNumericFields - returns standard stats for a list of numeric fields
- getStatsForSingleNumericField - returns standard stats for a single numeric field
- getStatsForSingleTextField - return stats for count and missing values for single text field

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/properties/PropertiesLoaderWeb.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/properties/PropertiesLoaderWeb.java
@@ -54,6 +54,7 @@ public class PropertiesLoaderWeb {
     public static final String LEAFLET_ATTRIBUTION_PROPERTY = "leaflet.attribution";
     public static final String TOP_LEFT_LOGO_IMAGE_PROPERTY = "top.left.logo.image";
     public static final String TOP_LEFT_LOGO_IMAGE_LINK_PROPERTY = "top.left.logo.image.link";
+    public static final String STATS_PROPERTY = "stats.fields";
  
     
     
@@ -87,8 +88,11 @@ public class PropertiesLoaderWeb {
 
     private static Properties serviceProperties = null;
     //Default values.
-    public static List<String> FACETS = Arrays.asList("domain", "content_type_norm", "type", "crawl_year", "status_code", "public_suffix"); 
+    public static List<String> FACETS = Arrays.asList("domain", "content_type_norm", "type", "crawl_year", "status_code", "public_suffix");
     public static String FIELDS=null;
+    public static List<String> STATS = Arrays.asList("content_length", "crawl_year", "content_text_length", "image_height", "image_width", "image_size",
+                                                    "links", "domain", "elements_used", "content_type",
+                                                    "content_language", "links_images", "type");
     
     //Default empty if not defined in properties
     public static  List<String> WORDCLOUD_STOPWORDS = new ArrayList<String>();
@@ -121,7 +125,8 @@ public class PropertiesLoaderWeb {
             isr.close();
 
             WAYBACK_SERVER =serviceProperties.getProperty(WAYBACK_SERVER_PROPERTY);
-            FACETS = Arrays.asList(getProperty(FACETS_PROPERTY, StringUtils.join(FACETS, ",")).split(", *"));		
+            FACETS = Arrays.asList(getProperty(FACETS_PROPERTY, StringUtils.join(FACETS, ",")).split(", *"));
+            STATS = Arrays.asList(getProperty(STATS_PROPERTY, StringUtils.join(STATS, ",")).split(", *"));
             WORDCLOUD_STOPWORDS = Arrays.asList(getProperty(WORDCLOUD_STOPWORDS_PROPERTY, StringUtils.join(WORDCLOUD_STOPWORDS, ",")).split(", *"));
             WEBAPP_PREFIX = serviceProperties.getProperty(WEBAPP_PREFIX_PROPERTY,"/solrwayback/"); //Default to /solrwayback/ if not defined
             OPENWAYBACK_SERVER = serviceProperties.getProperty(OPENWAYBACK_SERVER_PROPERTY);
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
@@ -4,6 +4,7 @@ import com.google.gson.Gson;
 import dk.kb.netarchivesuite.solrwayback.properties.PropertiesLoaderWeb;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.solr.client.solrj.SolrQuery;
+import org.apache.solr.client.solrj.SolrServerException;
 import org.apache.solr.client.solrj.response.QueryResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -91,11 +92,15 @@ public class SolrStats {
                 }
             }
 
-            QueryResponse response = NetarchiveSolrClient.query(solrQuery, true);
-            Gson gson = new Gson();
-            String stats = gson.toJson(response.getFieldStatsInfo().values());
+            try {
+                QueryResponse response = NetarchiveSolrClient.query(solrQuery, true);
+                Gson gson = new Gson();
+                String stats = gson.toJson(response.getFieldStatsInfo().values());
+                return stats;
+            } catch (Exception e){
+                throw new IllegalArgumentException("Percentiles have to be in range [0-100].");
+            }
 
-            return stats;
         }
     }
 
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
@@ -1,10 +1,14 @@
 package dk.kb.netarchivesuite.solrwayback.solr;
 
 import com.google.gson.Gson;
+import dk.kb.netarchivesuite.solrwayback.properties.PropertiesLoaderWeb;
 import org.apache.solr.client.solrj.SolrQuery;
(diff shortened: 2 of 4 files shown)


src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Get solr stats for single text field. Only return stats on count and missing values.
+     * @param query to generate stats for.
+     * @param field to return stats for.
+     * @return stats for count and missing as a JSON string.
+     */
+    public static String getStatsForSingleTextField(String query, String field){
+        SolrQuery solrQuery = new SolrQuery();
+        // Get all stats for input field
+        solrQuery.setQuery(query).addGetFieldStatistics("{!count=true missing=true}" + field);
```
tokee, line 72: `cardinality` is the HyperLogLog-thing. Shouldn't it be enabled?

src/test/java/dk/kb/netarchivesuite/solrwayback/SolrStatsTest.java
```diff
@@ -0,0 +1,147 @@
...
+        Assert.assertEquals(2003.0, entryAsJsonObject.get("min").getAsDouble(), 0);
+        Assert.assertEquals(2003.0, entryAsJsonObject.get("max").getAsDouble(), 0);
+        Assert.assertEquals(18027.0, entryAsJsonObject.get("sum").getAsDouble(), 0);
+        Assert.assertEquals(9, entryAsJsonObject.get("count").getAsInt());
+        Assert.assertEquals(0, entryAsJsonObject.get("missing").getAsInt());
+        Assert.assertEquals(2003.0, entryAsJsonObject.get("mean").getAsDouble(), 0);
+    }
+
+    @Test
+    public void multipleNumericFieldStatsTest(){
```
tokee, line 113: Maybe check that there is a sub-JSON for each requested field? (Just existende - no need to check all details for all fields)

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Get stats through the solr <a href="https://solr.apache.org/guide/8_11/the-stats-component.html">stats component</a>.
+ * Methods in this class
+ */
+public class SolrStats {
```
tokee, line 16: Consider adding support for `percentiles` - it is quite a nice feature for e.g. looking at image sizes.

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
```diff
@@ -0,0 +1,80 @@
...
+        return stats;
+    }
+
+    /**
+     * Get solr stats for single text field. Only return stats on count and missing values.
+     * @param query to generate stats for.
+     * @param field to return stats for.
+     * @return stats for count and missing as a JSON string.
+     */
+    public static String getStatsForSingleTextField(String query, String field){
```
tokee, line 69: Is it necessary to split requestsin numeric vs. text? What happens if e.g.. `avg` is requested from text field?

Alternative: As the valid text fields are known, they could be programmatically prepended with {!count=true missing=true}.

tokee (changes requested):
Nothing wrong with your code, it's all just suggestions/discussions.

tokee:
Extra idea: A method for requesting which fields it is possible to get stats for. For use by the frontend.

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
VictorHarbo: I have added a method for this now. Let me know what you think.

VictorHarbo:
New check-in:

- The three methods from before have been combined into the generic `getStatsForFields`. 
- A property called `STATS` has been created.
- A new method to extract percentile stats have been created.
- A method to show which fields have the stats component enabled.

Furthermore @tokee and I discussed and decided, that calculation of cardinality is not included in this first implementation as it requires differentiation between text and numeric fields.

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Get standard solr stats for all fields given.
+     * The solr documentation defines the standard stats <a href="https://solr.apache.org/guide/8_11/the-stats-component.html">here</a>
+     * @param query   to generate stats for.
+     * @param filters
+     * @param fields  to return stats for.
+     * @return all standard stats for all fields from query as a JSON string.
+     */
+    public static String getStatsForFields(String query, List<String> filters, List<String> fields){
```
tokee, line 33: You forgot to use the provided filters :-)

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
```diff
@@ -0,0 +1,92 @@
...
+        SolrQuery solrQuery = new SolrQuery();
+        solrQuery.setQuery(query);
+
+        for (String field: fields) {
+            if (PropertiesLoaderWeb.STATS.contains(field)){
+                solrQuery.setGetFieldStatistics(field);
+            } else {
+                log.warn("Stats can not be shown for field: " + field + " as it is not present in properties.");
```
tokee, line 41: As discussed this should throw an Exception instead, so that the user knows it was a (partially) invalid request.

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
```diff
@@ -0,0 +1,92 @@
...
+     * The solr documentation defines the standard stats <a href="https://solr.apache.org/guide/8_11/the-stats-component.html">here</a>
+     * @param query   to generate stats for.
+     * @param filters
+     * @param fields  to return stats for.
+     * @return all standard stats for all fields from query as a JSON string.
+     */
+    public static String getStatsForFields(String query, List<String> filters, List<String> fields){
+        SolrQuery solrQuery = new SolrQuery();
+        solrQuery.setQuery(query);
+
```
tokee, line 36: Maybe add a check for zero fields and either throw an Exception or assign some default fields?

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param query to generate stats for.
+     * @param percentiles to extract values for.
+     * @param fields to return percentiles for.
+     * @return percentiles for specified fields as a JSON string.
+     */
+    public static String getPercentilesForFields(String query, List<String> percentiles, List<String> fields){
+        SolrQuery solrQuery = new SolrQuery();
+        solrQuery.setQuery(query);
+
+        String allPercentiles = StringUtils.join(percentiles, ",");
```
tokee, line 62: Sorry, but this needs to be properly parsed to doubles first: https://xkcd.com/327/

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Get stats through the solr <a href="https://solr.apache.org/guide/8_11/the-stats-component.html">stats component</a>.
+ * Methods in this class
+ */
+public class SolrStats {
+    private static final Logger log = LoggerFactory.getLogger(SolrStats.class);
+    public static final List<String> interestingNumericFields =  Arrays.asList("content_length", "crawl_year", "content_text_length", "image_height", "image_width", "image_size");
+    final List<String> interestingTextFields = Arrays.asList("links", "domain", "elements_used", "content_type",
+                                                                "content_language", "links_images", "type");
+    final List<String> otherNumericFields = Arrays.asList("score", "status_code", "source_file_offset", "_version_", "wayback_date");
```
tokee, line 23: These are not currently used. Should they be default fields if none are provided?

src/main/java/dk/kb/netarchivesuite/solrwayback/properties/PropertiesLoaderWeb.java
```diff
@@ -87,8 +88,11 @@ public class PropertiesLoaderWeb {
 
     private static Properties serviceProperties = null;
     //Default values.
-    public static List<String> FACETS = Arrays.asList("domain", "content_type_norm", "type", "crawl_year", "status_code", "public_suffix"); 
+    public static List<String> FACETS = Arrays.asList("domain", "content_type_norm", "type", "crawl_year", "status_code", "public_suffix");
     public static String FIELDS=null;
+    public static List<String> STATS = Arrays.asList("content_length", "crawl_year", "content_text_length", "image_height", "image_width", "image_size",
```
tokee, line 93: These should also be added to the sample property file `src/test/resources/properties/solrwaybackweb.properties`.

tokee (changes requested):
Minor suggestions.

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
VictorHarbo: They were leftovers from earlier experimenting.

src/main/java/dk/kb/netarchivesuite/solrwayback/solr/SolrStats.java
tokee: Looks good!

VictorHarbo:
I think this does the trick. However I am not quite sure, that I am catching that SolrServerException correctly.

tokee (approved):

Issue #333, opened 2023-04-20
Handle non-standard HTTP header in warc-parser


thomasegense:
The WARC-parser will give error for content-length field in HTTP header.HTTP/1.1 200 OK
In this example tab or too many whitespaces after Content-Length

Server: BigIP
Content-Type: text/html; charset=utf-8
Accept-Ranges: bytes
Connection: close
Date: Tue, 01 Mar 2022 14:48:02 GMT
Age:     866
Content-Length:       5155
X-Frame-Options: DENY

thomasegense:
PR sent to @tokee

thomasegense:
Merged

Issue #332, opened 2023-04-20
How to handle OOM during export?


tokee:
If an OOM happens during export, there is currently no way to signal this to the client. The stream will end with 
```
<ns1:XMLFault xmlns:ns1="http://cxf.apache.org/bindings/xformat"><ns1:faultstring xmlns:ns1="http://cxf.apache.org/bindings/xformat">java.lang.OutOfMemoryError: Java heap space</ns1:faultstring></ns1:XMLFault>
```
but that is not immediately visible when exporting as `warc.gz` (calling `tail -c 1000 exported.warc.gz | strings` will show it).

Could we introduce some sort of download tracking? Maybe the client could provide a unique tracking ID to the export call and the GUI could busy-poll SolrWayback for status on that export?

PR #334, opened 2023-04-20
warc.gz export problem extras

### LINKED ISSUE #322: Export as zip-files sometimes gives corrupt zip-files
It has been reported to happen often for warc.gz export.  
The theory is that it is the last entry, that is missing correct zip-ending.

The export is called directly with wget. Here is an example, however this is a too large extraction for test. It will take many hours.

...../solrwayback/services/export/warcExpanded?query=arc_job:383960%20AND%20domain:bronderslev.dk&gzip=true

tokee:
Export to `warc.gz` caused truncated files in production: #332 

The cause seems to be #332 + #333 but during the investingation some minor enhancements were introduced to the code base, primarily Exception handling for export. This pull request contains those.

This closes #322

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ The default number of 20 results on each page can be changed with this new prope
 
 search.pagination=20
 
+Bugfix: The property key for expanded warc limit was the same as for non-expanded
 
 
 4.4.0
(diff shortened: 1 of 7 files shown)


thomasegense (approved):
Testing shows export still works. Hard to test if it fixes the out of memory error

thomasegense (commented):
OK!

PR #335, opened 2023-04-20
Fixed parsing of content-length in HTTP header if tab/white spaces.

thomasegense:
Added unittest and a new WARC-file to test resources.

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
@@ -266,7 +266,7 @@ public class WarcParser extends  ArcWarcFileParserAbstract {
     else if (headerLine.toLowerCase().startsWith("transfer-encoding:")) {                                      
       String transferEncoding=headerLine.substring(18).trim();
       //log.debug("transfer-encoding:"+transferEncoding);
-      if (transferEncoding.equalsIgnoreCase("chunked")) {
+      if (transferEncoding.indexOf("chunked") > 0) { //multivalued
         warcEntry.setChunked(true);
       }
     }    
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
@@ -61,13 +61,11 @@ public class WarcParser extends  ArcWarcFileParserAbstract {
     warcEntry.setFormat(ArcEntry.FORMAT.WARC);
     warcEntry.setSource(arcSource);
     warcEntry.setOffset(warcEntryPosition);
-
+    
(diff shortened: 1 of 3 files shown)


src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
```diff
@@ -267,7 +265,7 @@ else if (headerLine.toLowerCase().startsWith("location:")) {
     }
     else if (headerLine.toLowerCase().startsWith("transfer-encoding:")) {                                      
       String transferEncoding=headerLine.substring(18).trim();
```
tokee, line 267: Why use `substring` here when `split` is used on line 255? Consistent solutions to the same problem makes the code easier to real. Also, the use of constants such as `18` makes it harder to determine if the code does what it is supposed to.

src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
```diff
@@ -254,7 +252,7 @@ private static void populateWarcSecondHeader(ArcEntry warcEntry, String headerLi
 
     }  //Content-Length: 31131
     else if (headerLine.toLowerCase().startsWith("content-length:")) {
-      String[] contentLine = headerLine.split(" ");
+      String[] contentLine = headerLine.split(":");
       long totalSize = Long.parseLong(contentLine[1].trim());
```
tokee, line 256: Under the assumption that the `content-length` follows [RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#page-31), this should work.

src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
```diff
@@ -267,7 +265,7 @@ else if (headerLine.toLowerCase().startsWith("location:")) {
     }
     else if (headerLine.toLowerCase().startsWith("transfer-encoding:")) {                                      
       String transferEncoding=headerLine.substring(18).trim();
-      log.debug("transfer-encoding:"+transferEncoding);
+      //log.debug("transfer-encoding:"+transferEncoding);
       if (transferEncoding.equalsIgnoreCase("chunked")) {
```
tokee, line 269: This should be a contains as the transfer-encoding can have more than one value: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding

tokee (commented):
Up to you if you want to change based on this review. Functionally it works.

src/main/java/dk/kb/netarchivesuite/solrwayback/parsers/WarcParser.java
thomasegense: Because transfer encoding can contain the ':' character as well.

PR #338, opened 2023-04-23
Harden export and playback against OOM

### LINKED ISSUE #337: Avoid full binary load
The current abstraction for (W)ARC entries in SolrWayback has support for optionally providing "binaries", meaning that the full binary content of a given WARC entry is loaded in memory. Although there is a maximum of 2GB due to `byte[]` size limit in Java, it is still problematic as the World Wide Web has an endless amount of edge cases: At the royal Danish Library we ran into problems with a 100MB webpage.

To harden against Out Of Memory we should
1. Remove the old "load binary" option
2. Stream binaries directly from the source, with a reasonable memory cache to avoid multiple identical requests in most cases
3. Provide a text-oriented option used whenever the content for a WARC entry needs to be parsed (e.g. extracting links from HTML pages) that has an upper bound on the number of extracted characters

tokee:
This is a major clean up of binary handling for `ArcEntry` and related classes. No binaries are loaded up front and wherever possible, `InputStream`s or `Reader`s are used to deliver the content. There are still a few places where it is necessary to load the binary on heap and for that a limit (default: 100MB) has been introduced on the size of the extracted binary.

This should makes it less likely that an Out Of Memory Error is thrown during export (specifically link resolving as part of expanded export) or playback of HTML, CSS and JavaScript content.

This closes #337.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@ Bugfix: The property key for expanded warc limit was the same as for non-expande
 
 Default log configuration will create and addition log file called solrwayback_error.log. This will only contain log lines of severity 'error'.  
 
+Hardening against Out Of Memory in export and playback of very large webpages
+
 4.4.0
 -----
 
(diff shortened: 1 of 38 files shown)


thomasegense (approved):
Everything working. 
Added changelog about new property
removed  spammy log debug about loading buffers for export

PR #339, opened 2023-04-24
218 deliver query stats

VictorHarbo:
Relates to #218 

I am not sure that the endpoint names are fitting. Let me know what you think.


thomasegense (changes requested):
There are some unittest issues.

Try run from IDEE and also from mvn.

thomasegense (approved):

PR #344, opened 2023-05-01
Github 299 - Remove search result capital letters

jorntx:
Remove all caps from titles in search result


thomasegense (approved):
Tested, works for all content types.

PR #345, opened 2023-05-01
342 cache invalidation

### LINKED ISSUE #342: CachingSolrClient, overflow
CachingSolrClient has an overflow issue. It is multiplied by 1000 and should not.
Also there are some logging that divides by 1000  that may be wrong-

tokee:
The conversion from seconds to milliseconds for max age in `CachingSolrClient` has an implicit cast from the given `int` to `long`, according to IntelliJ IDEA. The added unit tests showed that was a lie: There were no cast and the `int` overflowed to negative, effectively disabling the cache.

This pull request adds unit tests for the cache module and fixed the overflow problem with an explicit promotion to `long`. This closes #342


thomasegense (approved):
seems to work.
Can never unittest too much it seems-

PR #346, opened 2023-05-01
github-300 wider search field

jorntx:
Make search input field wider on wide screens


thomasegense (approved):
Working, but not please with the magic numbers (1721px).

PR #358, opened 2023-05-13
Lower sample call for WARC indexing in the bundle from 6 to 2 threads

### LINKED ISSUE #357: Change default number of index threads in the bundle
The README.md for bundle use stated 6 threads as default. Each thread takes 1GB of RAM. This is too much for testing. Maybe 2 threads?

tokee:
The old sample with 6 threads required 6GB of RAM for indexing alone. As people tend to copy and paste instead of reading all the text, this caused problems for smaller machines (and virtual machines that tend to be more RAM limited).

This pull request is simply an update of `README.md`.

This closes #357


thomasegense (approved):
OK

PR #359, opened 2023-05-13
Switch from wget to curl in the index scripts

tokee:
`wget` is used in the bundle indexing scripts for pinging Solr and for triggering a Solr commit. `wget` is not readily available on macOS, [but curl is](https://everything.curl.dev/get/macos). This pull request switches from `wget` to `curl` in the scripts. There is no change in functionality.


thomasegense (approved):
OK

PR #378, opened 2023-06-27
Make curl and wget interchangeable and optional in the bundle

tokee:
This pull request changes the `warc-indexer.sh` to seamlessly choose between `curl` or `wget` depending on which is available. If none are available, a warning is logged and `ping` & `commit` are not called - base indexing should work without `curl` and `wget`.

### DIFF
--- a/src/bundle/indexing/warc-indexer.sh
+++ b/src/bundle/indexing/warc-indexer.sh
@@ -8,6 +8,8 @@
 #
 # 2021-06-07: Initial script
 # 2021-06-08: Status & tmp folder moved away from the WARC location
+# ... Missing updates
+# 2023-06-27: Both curl and wget can be used
 #
 
 ###############################################################################
@@ -160,6 +162,13 @@ check_parameters() {
     fi
 }
 
+# Either wget or curl must be present
+check_tools() {
+    if [[ -z $(which curl) && -z $(which wget) ]]; then
+        >&2 echo "Warning: Neither 'curl' nor 'wget' is present. ping and commit will be disabled"
+    fi
+}
+
 ################################################################################
 # FUNCTIONS
 ################################################################################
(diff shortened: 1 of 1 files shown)


src/bundle/indexing/warc-indexer.sh
```diff
@@ -160,6 +162,13 @@ check_parameters() {
     fi
 }
 
+# Either wget or curl must be present
+check_tools() {
+    if [[ -z $(which curl) && -z $(which wget) ]]; then
+        >&2 echo "Warning: Neither 'curl' nor 'wget' is present. ping and commit will be disabled"
```
VictorHarbo, line 168: This warning is never shown on mac

src/bundle/indexing/warc-indexer.sh
```diff
@@ -256,7 +265,15 @@ check_solr() {
...
     fi
     echo " - Checking if Solr is running"
-    curl  -s "${SOLR_URL}/admin/ping" > /dev/null
+
+    if [[ ! -z $(which curl) ]]; then
+        curl  -s "${SOLR_URL}/admin/ping" > /dev/null
+    elif [[ ! -z $(which wget) ]]; then
+        wget  -nv -O- "${SOLR_URL}/admin/ping" > /dev/null
+    else
+        >&2 echo "Warning: Unable to check for Solr availability as neither 'curl' nor 'wget' is present"
```
VictorHarbo, line 274: However, this one gets shown correctly

src/bundle/indexing/warc-indexer.sh
```diff
@@ -271,7 +288,19 @@ commit() {
...
+    
+    if [[ ! -z $(which curl) ]]; then
+        curl  -s "$COMMIT_URL" > /dev/null
+    elif [[ ! -z $(which wget) ]]; then
+        wget  -nv -O- "$COMMIT_URL"  > /dev/null
+    else
+        >&2 echo "Warning: Unable to trigger Solr commit as neither 'curl' nor 'wget' is present"
+        >&2 echo "Either wait 10 minutes or visit the following URL in a browser to trigger commit:"
+        >&2 echo "$COMMIT_URL"
+        true
```
VictorHarbo, line 301: This also works!

VictorHarbo (changes requested):
I have tested on a Mac. I tested as you suggested Toke, by changing curl and wget to nejCurl and nejWget. The `check_tools()`-method does not throw the warning to the screen.

VictorHarbo:
Also: Should we submit these changes to the original [warc-indexer ](https://github.com/ukwa/webarchive-discovery) as other users might encounter the same dilemma of tooling on different setups?

src/bundle/indexing/warc-indexer.sh
tokee: It's not enough to write the methods? We have to call them too?

VictorHarbo:
LGTM!

PR #380, opened 2023-06-27
Avoid query warnings for datetime range queries

tokee:
The query sanity check raises a warning for valid datetime ranges such as `crawl_time:[2023-12-99T12:34:56Z TO *]`. This pull request disables false warnings of that type.


thomasegense:
LGTM

PR #385, opened 2023-06-28
Regexp based file resolver

### LINKED ISSUE #384: Add regexp based WARC resolver
The [FileMovedMappingResolver](https://github.com/netarchivesuite/solrwayback/blob/master/src/main/java/dk/kb/netarchivesuite/solrwayback/interfaces/FileMovedMappingResolver.java) allows for arbitrary reorganization of WARC file locations by using a map from WARC names to folders.

A common pattern for reorganization is to move a full subtree to another location, e.g.`/home/harvest/warcs/<WARC-files>` → `/netarchive/warcs/<WARC-files>`. Instead of maintaining a fill map of all WARCs, a regexp based search-replace of the location could be used for this scenario: With the previous example, a regexp with `/home/harvest/(.*)` → `/netarchive/$1` would suffice.

Such a resolver should be added to SolrWayback. Naming suggestion is `FileMovedRegexpResolver` to match `FileMovedMappingResolver`.

tokee:
Basic implementation of a regexp bases search/replace for file paths to support the case of moving a full subtree of folders with WARCs to another main file location.

This closed #384.

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/interfaces/FileMovedRegexpResolver.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/interfaces/FileMovedRegexpResolver.java
@@ -1,0 +1,83 @@
+package dk.kb.netarchivesuite.solrwayback.interfaces;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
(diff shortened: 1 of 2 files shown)


thomasegense:
Can you add the fileresolver to solrwayback.properties ? Commented with a working example and few lines documentation

tokee:
This has been superceded by #387 which now has the same files as in this pull request. I suggest we close #385 without merging and focus on #387 instead.

thomasegense:
Closed, will look at #385 instead

Issue #245, opened 2022-09-09
Add ZIP as export option


tokee:
At the Royal Danish Library we have had multiple requests for exporting specific resources for a query, e.g. "All PDFs from domain X". Currently is can be done by exporting to WARC and then use a tool to convert to individual files. We should add the option of exporting to a ZIP for easier use.

There are some problems with this approach:
 * What should the file structure be? `domain/url-path/file` seems logical enough, but what if the URL is `http://example.com/foo?bar=87?type=pdf`?
 * How to handle name-duplicates? We could choose to use the hash as part of the file path, but that is a hassle when there are no or very few duplicates. Maybe store the first instance directly and subsequent instances with their hashes appended to the file name?

We should check how other tools handles that.

PR #390, opened 2023-07-24
Zip export.

### LINKED ISSUE #382: Batch content export
At the Royal Danish Library there has been multiple requests for en masse exporting raw archive content, e.g. unmodified HTML, images or PDFs. The current exporter only supports WARC for this and for some researchers they can be cumbersome to work with.

SolrWayback should have an export option for a more common container format, where 64-bit zip is the obvious candidate as "all" platforms supports it out of the box.

The big question is how to handle naming for non-WARC export. Two options comes to mind:

1. Best effort ala `timestamp/Filename_cleaned_of_non-ASCII_spaces_and_similar.ext`
2. `timestamp_hash.exe` with a `metadata.txt` which contains `timestamp, hash, WARC-file, WARC-offset, URL`

VictorHarbo:
Extraction of content from WARC files by mimetype

Closes #382 and #245


src/main/java/dk/kb/netarchivesuite/solrwayback/export/StreamingRawZipExport.java
```diff
@@ -0,0 +1,145 @@
...
+        output.close();
+        log.info("Streamed {} warc entries with the contentType: '{}'.", streamedDocs, contentType);
+    }
+
+    /**
+     * Normalize content type.
+     * @param contentType string to normalize.
+     * @return            normalized contentType string.
+     */
+    private String normalizeContentType(String contentType) {
```
thomasegense, line 65: Not used anymore

Issue #298, opened 2022-12-01
Applied Facets: you can select the same facet several times


Klindten:
Task: Shouldn't it be checked if a facet was already selected so that the list doesn't just continue if you select the same facet several times?
![image](https://user-images.githubusercontent.com/74599011/205067385-d693d656-b0de-4bb4-a2a1-d55ba7267107.png)

thomasegense:
Assigned to @jorntx .
I do not know if this is an easy fix.

jesperlauridsen:
Merged into master, closing the issue. Will be in the next release.

PR #393, opened 2023-08-07
github-298: facet addable multiple times

jesperlauridsen:
Should take care of the issue: https://github.com/netarchivesuite/solrwayback/issues/298

### DIFF
--- a/src/js/src/components/SearchFacetOptions.vue
+++ b/src/js/src/components/SearchFacetOptions.vue
@@ -83,7 +83,7 @@ export default {
       this.searchAppliedFacets.forEach((facet) => {
         facet === newFacet ? facetAllreadyApplied = true : null
       })
-      if(!facetAllreadyApplied === true) {
+      if(!facetAllreadyApplied) {
         this.updateSolrSettingOffset(0)
         this.addToSearchAppliedFacets(newFacet)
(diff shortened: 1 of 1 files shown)


src/js/src/components/SearchFacetOptions.vue
```diff
@@ -78,10 +78,16 @@ export default {
...
     applyFacet(facetCategory, facet) {
+      let facetAllreadyApplied = false
       let newFacet = '&fq=' + facetCategory + ':"' + facet + '"'
-      this.updateSolrSettingOffset(0)
-      this.addToSearchAppliedFacets(newFacet)
-      this.$_pushSearchHistory('Search', this.query, this.searchAppliedFacets, this.solrSettings)
+      this.searchAppliedFacets.forEach((facet) => {
+        facet === newFacet ? facetAllreadyApplied = true : null
+      })
+      if(!facetAllreadyApplied === true) {
```
jorntx, line 86: Could shorten this to `if(!facetAllreadyApplied) {`

jorntx (approved):
Looks good and works as well. Had one small comment which doesn't change a whole lot:)

I like your branch naming and I think we should adobt this approach going forward and maybe throw in an issue number.

src/js/src/components/SearchFacetOptions.vue
jesperlauridsen: Aaah! Yes, i'll fix that!

PR #396, opened 2023-08-09
395 Fix bug when combining filters

### LINKED ISSUE #395: SolrUtils.combineFilterQueries combines wrong
`SolrUtils.combineFilterQueries` combines using ` AND `, which leads to changed filter behaviour us the input is e.g. `foo:bar OR bar:zoom` and `boom:moo`, which becomes `(foo:bar OR bar:zoom AND boom:moo)`, where it should have been `(foo:bar OR bar:zoom) AND (boom:moo)`. The last parentheses pair is not needed but also does not hurt.

Replacing
```
    .collect(Collectors.joining(" AND ", "(", ")"));
```
with
```
    .collect(Collectors.joining(") AND (", "(", ")"));
```
should fix it.

VictorHarbo:
closes #395


tokee (approved):
Looks good!

PR #399, opened 2023-08-09
353/feature/config resolve

### LINKED ISSUE #353: Better resolving of configurations
When using the SolrWayback bundle, the user has to copy (and adjust) the default property files to the home folder. The concept of "home folder" is not very transparent on Windows (or OS-X, depending on user experience). SolrWayback bundle should be able to find configurations in the solrwayback folder.

tokee:
Adds more flexible resolving of the config files. The use case is for the SolrWayback bundle, where the config files can now be placed directly in the bundle folder. It it also be possible to place the config files in the root of the project folder with the source code, when running under jetty.

If the location of the two property files are not specified in the WAR context, they are sought resolved from the container (e.g. tomcat or jetty) root. If not there, the parent folder is checked and so forth. If that does not yield results, the current folder as well as the user's home folder is checked in a similar manner.

Testing this properly is cumbersome as it requires at least

 * Placing a config file in the root of the project folder and running with jetty
 * Placing a config file in the root a tomcat folder where the WAR is deployed
 * Specifying a config file location in the context (i.e. `solrwayback.xml`) for SolrWayback using tomcat and checking that this is used instead of another config placed directly in the tomcat folder. See the `webapp/META-INF/context.xml` for how to specify the location explicitly

This closes #353

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/listeners/InitializationContextListener.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/listeners/InitializationContextListener.java
@@ -34,6 +34,10 @@ public class InitializationContextListener implements ServletContextListener {
             version = props.getProperty("APPLICATION.VERSION");
             PropertiesLoaderWeb.SOLRWAYBACK_VERSION = version;
 
+            // Remove leading "/"
+            webAppContext = webAppContext.startsWith("/") && !webAppContext.substring(1).contains("/") ?
+                    webAppContext.substring(1) :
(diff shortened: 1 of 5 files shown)


VictorHarbo (changes requested):
I've tested all three config placements and found one issue.
I can't seem to get the context.xml configuration working. When I add a path to get config from it does not update when i restart the tomcat. If i then redeploy the webapp, the contenxt.xml gets overwritten.

VictorHarbo (approved):
LGTM!

PR #400, opened 2023-08-10
feat: add `Dockerfile` template

trym-b:
# Motivation

More and more applications are being run natively in containers, for a variety of reasons. This commit is the initial attempt to provide containerized support for `SolrWayback`.

# Usage

To get going, simply run the following command:

```bash
docker build . --tag solrwayback
docker run --publish 8080:8080 --publish 8983:8983  --volume \
 <path/to/WARCs>:/host_dir --tty --interactive solrwayback bash
```
where `<path/to/WARCs>` only contains `WARC` files and directories.

For more details, please refer to the comments in the `Dockerfile`.

# Implementation

The `Dockerfile` was created by following the instructions in the top-level readme. In addition to this, a few verification steps were added to ensure that the container works as expected.

A simple test that uses `docker` to build the container has been added.

# Drawbacks

## No proper indexing test

The added test does not verify that indexing works as expected. This should be done by adding some `WARC` files as test data, but this is outside the scope of this commit.

## No automatic update to latest release

Whenever someone releases a new version of `SolrWayback`, there is no reminder or automatic failure if the relevant people forget to update the `Dockerfile`.

# Future work

## Index preservation

The `Dockerfile` does currently not preserve the created index, so it needs to be manually copied out of the container in order to preserve it. With the latest released `SolrWayback` bundle, the index can be found here:
`unpacked-bundle/solrwayback_package_4.4.2/solr-7.7.3/server/solr/ configsets/netarchivebuilder/netarchivebuilder_data/index/`

## Custom configuration of `properties`

There is currently no way of using your own `solrwayback.properties` and `solrwaybackweb.properties`, which is essential for using the correct branding.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,14 @@
+name: "Test"
+
+on:
+  push:
+
+jobs:
+  build-docker-image:
+    name: Build Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Build SolrWayback Docker image
+        run: docker build --tag solrwayback .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,61 @@
+# This dockerfile sets up the SolrWayback bundle and attempts to run both Solr and Tomcat.
+# To build the image, run:
+# docker build . --tag solrwayback
+
+# To run SolrWayback, you need to launch it with the following parameters
+# docker run --publish 8080:8080 --publish 8983:8983  --volume <path/to/WARCs>:/host_dir --tty --interactive solrwayback bash
+# where <path/to/WARCs> is a file path that only contains WARC files and directories.
+
+# When the container is running, run the following commands to start Solr and Tomcat:
+# export SOLRWAYBACK_VERSION=4.4.2
+# export APACHE_TOMCAT_VERSION=8.5.60
+# export SOLR_VERSION=7.7.3
+# ./unpacked-bundle/solrwayback_package_$SOLRWAYBACK_VERSION/solr-$SOLR_VERSION/bin/solr start
+# ./unpacked-bundle/solrwayback_package_$SOLRWAYBACK_VERSION/apache-tomcat-$APACHE_TOMCAT_VERSION/bin/startup.sh
+
+# You should now verify that the following links works with a browser:
+# http://localhost:8080/solrwayback/
+# http://localhost:8983/solr/#/
+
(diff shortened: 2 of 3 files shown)


thomasegense:
I will try have a look of this later this week. It is important you can add your own WARC-files as you have documented.  There is no need to have some default WARC-files included. If users want to try SolrWayback they can just go to live web and
see one of the many running installations with real data. So this is all good.

How do you see the solrwayback.log and solrwayback_error.log files?  Can you open a terminal inside the docker? Then you should also be able to edit the two property files if needed?

I tried run the github workflow on the branch and it passed. For some reason the old docker example (I deleted) would make github fail workflow runs, so we never know if it could even compile. I never understood why github would care about a docker file in the root folder, but this seems to cause no problems.

Also using the SolrWayback Bundle for installing everything makes everything so much easier. The old docker had a full git checkout of solrwayback that was build and WAR-file deployed into tomcat. But I guess you can just have a git checkout outside the docker container and then easy copy the WAR-file into the docker tomcat?  This could be useful for people trying to create PR etc.

trym-b:
> How do you see the solrwayback.log and solrwayback_error.log files? Can you open a terminal inside the docker? Then you should also be able to edit the two property files if needed?

If you follow the instructions provided in `Dockerfile` you can easily inspect the log files by either using `less`, `cat`, `nano` or something else. I am not currently sure where those lives exactly, but if they appear in the bundle itself you can find them in `~/unpacked-bundle/solrwayback_package_${SOLRWAYBACK_VERSION}/` somewhere.

> But I guess you can just have a git checkout outside the docker container and then easy copy the WAR-file into the docker tomcat? 

My approach is to not deal with `git` at all, as I do not have a need for it. I simply use the latest bundle that has been released.

If you take a look at the `Dockerfile` you will find that I added some lines that explain how you are supposed to get `WARC` files into the container:
```
# To run SolrWayback, you need to launch it with the following parameters
# docker run --publish 8080:8080 --publish 8983:8983  --volume <path/to/WARCs>:/host_dir --tty --interactive solrwayback bash
# where <path/to/WARCs> is a file path that only contains WARC files and directories.
```
You could manually use `COPY` to copy the files into the container, but that would increase the size of the container/image, so instead I recommend that you `mount` the relevant directory so that you can just use the files directly.

This is of course not a silver bullet, as you might have a different use-case where it makes sense to copy files instead of mounting them. But this is also not the point of this pull request, it is simply for having a sample that you can start modifying. For example, there is a use-case for mounting a different directory for exporting the index that is created, which might be added as a follow-up.

thomasegense (commented):
Working as expected. 
Thanks.

PR #403, opened 2023-08-10
Wok 667 double spinner on ngram

jesperlauridsen:
Should remove the double spinner


jorntx (approved):
Much better than my solution - thanks for correcting that and now we have corresponding spinners across the toolbox. Tested and verified.

PR #406, opened 2023-08-11
CTRL + click on facet to open in new tab

### LINKED ISSUE #404: Facets. Make it possible to open new tab in browser when CTRL+clicking or Right click>"open link in new tab/window"
When CTRL+clicking (or Right click  "open link in new tab/window" under the facets you navigate to the linked page instead of opening a new tab. 

In some cases it´s great to start loading multiple results under facets using tabs,

jesperlauridsen:
Added the feature, while keeping the normal interaction without CTRL.

closes #404

### DIFF
--- a/src/js/src/components/SearchFacetOptions.vue
+++ b/src/js/src/components/SearchFacetOptions.vue
@@ -84,12 +84,11 @@ export default {
         facet === newFacet ? facetAllreadyApplied = true : null
       })
       if(!facetAllreadyApplied) {
-          if(event.ctrlKey === true) {
+        if(event.ctrlKey || event.metaKey) {
             const localSolrSettings = JSON.parse(JSON.stringify(this.solrSettings))
             const localFacets = [...this.searchAppliedFacets]
             localFacets.push(newFacet)
             localSolrSettings.offset = 0
-            console.log(this.$_getResolvedUrl('Search', this.query, localFacets, localSolrSettings))
             window.open(this.$_getResolvedUrl('Search', this.query, localFacets, localSolrSettings).href, '_blank')
           }
           else {
@@ -98,7 +97,8 @@ export default {
             this.$_pushSearchHistory('Search', this.query, this.searchAppliedFacets, this.solrSettings)
         }
       }
-      else if(facetAllreadyApplied && event.ctrlKey === true) {
+      else if(facetAllreadyApplied && event.ctrlKey ||
+              facetAllReadyApplied && event.metaKey) {
         window.open(this.$_getResolvedUrl('Search', this.query, this.searchAppliedFacets, this.solrSettings).href, '_blank')
       }
     },
--- a/src/js/src/components/SearchFacetOptions.vue
+++ b/src/js/src/components/SearchFacetOptions.vue
@@ -12,7 +12,7 @@
         <div v-for="(facet, facetIndex) in facetCategory[1]"
              :key="facetIndex"
              :class="facetIndex % 2 === 0 ? 'facetItem' : 'facetCount'"
-             @click="facetIndex % 2 === 0 ? applyFacet(facetCategory[0], facet) : null">
+             @click="facetIndex % 2 === 0 ? applyFacet(facetCategory[0], facet, $event) : null">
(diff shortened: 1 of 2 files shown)


src/js/src/components/SearchFacetOptions.vue
```diff
@@ -77,17 +77,30 @@ export default {
...
       let newFacet = '&fq=' + facetCategory + ':"' + facet + '"'
       this.searchAppliedFacets.forEach((facet) => {
         facet === newFacet ? facetAllreadyApplied = true : null
       })
       if(!facetAllreadyApplied) {
-        this.updateSolrSettingOffset(0)
-        this.addToSearchAppliedFacets(newFacet)
-        this.$_pushSearchHistory('Search', this.query, this.searchAppliedFacets, this.solrSettings)
-    }
+          if(event.ctrlKey === true) {
```
jorntx, line 87: Good to make use of the 'event' to handle the ctrlKey issue but you need to handle 'command' to include Mac-user too. So I propose the following correction to hande the Apple Command key:

`if(event.ctrlKey || event.metaKey) {` // which also takes my comment below into account:) 

Instead of saying anything about the evaluation of 'true' you have here I will just use three of these ...

src/js/src/components/SearchFacetOptions.vue
```diff
@@ -77,17 +77,30 @@ export default {
...
+            const localSolrSettings = JSON.parse(JSON.stringify(this.solrSettings))
+            const localFacets = [...this.searchAppliedFacets]
+            localFacets.push(newFacet)
+            localSolrSettings.offset = 0
+            console.log(this.$_getResolvedUrl('Search', this.query, localFacets, localSolrSettings))
```
jorntx, line 92: We don't want this console.log in production

src/js/src/components/SearchFacetOptions.vue
```diff
@@ -77,17 +77,30 @@ export default {
...
+            window.open(this.$_getResolvedUrl('Search', this.query, localFacets, localSolrSettings).href, '_blank')
+          }
+          else {
+            this.updateSolrSettingOffset(0)
+            this.addToSearchAppliedFacets(newFacet)
+            this.$_pushSearchHistory('Search', this.query, this.searchAppliedFacets, this.solrSettings)
+        }
+      }
+      else if(facetAllreadyApplied && event.ctrlKey === true) {
```
jorntx, line 101: Instead of saying anything about the evaluation of 'true' here I will just use three os these ...

src/js/src/mixins/HistoryRoutingUtils.js
```diff
@@ -24,6 +24,17 @@ export default {
         currentParams.facets !== appliedFacets.join('') ||
         currentParams.offset !== solrSettings.offset ||
         currentParams.sort !== solrSettings.sort)
+    },
+    $_getResolvedUrl(destination, query, appliedFacets, solrSettings) {
+      const route = this.$router.resolve({name: destination, query: { query: query,
```
jorntx, line 29: So good practice you utilised $router.resolve here.

jorntx (changes requested):
Good solution, tested but failed on Apple devices so I have proposed a fix for this.

Besides that only minor comments and some praise.

src/js/src/components/SearchFacetOptions.vue
jesperlauridsen: I must admit i totally forgot about mac keyboards. Damn, good catch. Fixed that now.

src/js/src/components/SearchFacetOptions.vue
jesperlauridsen: Missed one, sorry.

src/js/src/components/SearchFacetOptions.vue
jesperlauridsen: Hah, i don' know why i do it. Old habbits die hard. Fixed.

src/js/src/mixins/HistoryRoutingUtils.js
jesperlauridsen: :+1:

PR #407, opened 2023-08-11
Set explicit encoding UTF-8 for indexing

### LINKED ISSUE #405: Set file encoding to UTF-8 when calling warc-indexer
During work with Docker, @trym-b discovered that [webarchive-discovery](https://github.com/ukwa/webarchive-discovery) used by the SolrWayback bundle needs to have the environment `file.encoding` set to `UTF8` to output UTF-8 for the Solr index. Ideally this should not be necessary, and [webarchive-discovery issue 311](https://github.com/ukwa/webarchive-discovery/issues/311) has been opened for `webarchive-discovery` to fix this.

In the meantime, the index-scripts in the SolrWayback bundle (`warc-indexer.sh` et al) should set
```
JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF8
```
to ensure proper encoding.

tokee:
The warc-indexer used for indexing WARCs to Solr does not specify UTF-8 as the output character encoding itself. While UTF-8 is default on many systems, it is still an assumption and if e.g. ASCII is the default, all non-ASCII characters will be replaced with an "unknown character" representation (typically a question mark in a black rhombus). This was triggered by Trym Brennes while running the bundle under Docker.

Setting file.encoding to UTF-8 is a workaround until the warc-indexer is changed to always output UTF-8. As the output is used for Solr (or Elasticsearch) which both supports UTF-8, this is the only encoding that makes sense.

This closes #405

### DIFF
--- a/src/bundle/indexing/batch_warcs1_folder.bat
+++ b/src/bundle/indexing/batch_warcs1_folder.bat
@@ -1,4 +1,4 @@
 cd /D "%~dp0"
 
-FOR /R warcs1 %%G IN (*.*) do  java -Xmx2048M -Djava.io.tmpdir=tika_tmp -jar warc-indexer-3.2.0-SNAPSHOT-jar-with-dependencies.jar -c config3.conf -s  "http://localhost:8983/solr/netarchivebuilder"  "%%G"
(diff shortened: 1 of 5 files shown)


VictorHarbo (approved):
LGTM!

Tested on ubuntu.

PR #410, opened 2023-08-15
Upgrade Java 1.8 → 11, Tomcat 8.5 → 9 and Solr 7 → 9

### LINKED ISSUE #331: Upgrade Java, Solr, Tomcat
It is about time to switch fully to java 11. This will also require upgrading the Tomcat+Solr versions in the bundle install.
Java 8-> 11
Tomcat 8->9 
Solr 7->9   (Running large installations can still keep running Solr 7, since indexing can be a major task)

Awating @tokee  Has validated the Solr9 schema and tested unitt-tests can still run with the embedded Solr9 server.

tokee:
This pull request upgrades Java, Tomcat & Solr. The Java and Tomcat bumps are near-trivial, but the Solr 7 → 9 bump introduces several changes. Besides updates to the code, the README for building a bundle (`src/bundle/README.md`) has been substantially extended. I suggest following that playbook and testing the produced bundle.

With the Solr upgrade comes a switch from Standalone to Cloud. This is not strictly necessary, but running in Cloud mode enables the use of Solr streaming, which has proven useful for flexible exports. There is near-zero penalty to running in Cloud, when using the bundle, as the only change from a user-perspective is the additional `-c` switch when starting Solr.

I have added both @thomasegense and @VictorHarbo as reviewers - we should test all the different features of SolrWayback due to the double Solr version jump.

SolrWayback _should_ still be backwards compatible with Solr 7. This is fairly easy to test by starting an existing Solr 7 installation instead of a new Solr 9.

This closes #331


thomasegense (approved):
Tested, and also backwards compatible to solr7 index.

Issue #329, opened 2023-03-27
Improve export speed


tokee:
The current experimental fields-exporter in solrWayback issues standard `cursorMark` requests against the underlying Solr.

Assume we have 100 Solr nodes and page 1000 documents at a time (a reasonable approximation of the index at the Royal Danish Library which has 143 nodes).

In a classic Solr Cloud setup each cursorMark request for 1000 documents will cause processing at each Solr node, where each node will return 1000 document-IDs and sort values, but only 10 will be used on average (1000 documents / 100 nodes = 10 documents/node). This means there is a lot of wasted processing for each request. A local test gave an export with high variance in speed, from 12-130 records/second and high load across the Solr Cloud.

If the Solr nodes were queried sequentially, each cursorMark request would involve only a single Solr node and all 1000 documents returned would be used. This was tested against the setup at [Netarkivet](https://www.kb.dk/find-materiale/samlinger/netarkivet) and the export speed here was ~1200 documents/second with load localized to a single Solr node. It would be quite feasible to query 1/4 of all nodes on parallel, which would increase export speed to ~30,000 documents/second in our not-so-hypothetical setup.

There is clearly much to gain by rewiring the SolrWayback export code. Two possibilities comes to mind:

## External shard requests
Introduce an endpoint listing the shards in the Solr setup and add a `shardID` parameter to the export endpoint.

* Pro: Very easy to implement in SolrWayback
* Con: Puts the burden of merging (and grouping if that is required) the result on the caller

## Internal shard requests
Add parallel explicit shard requests to the existing code, taking care to support grouping. Basically a mimic of [Solr streaming](https://solr.apache.org/guide/8_8/streaming-expressions.html) with the crucial difference that export of stored fields is possible.

* Pro: Makes field export performant without any extra work for the end user
* Con: Not trivial to implement, needs consideration of Solr Cloud topology

tokee:
While there is some bookkeeping when implementing the internal shard requests, the overall logistics is reasonably straight forward:

1. Use [CLUSTERSTATUS](https://solr.apache.org/guide/7_0/collections-api.html) to get the full topology
2. Get the true collection IDs by checking the alias section
3. For each true collection, get all shard IDs
4. For all shard IDs, create a streaming document provider, either by specifying the shard ID in the request or directly connecting to the `base_url` stated in the topology map
5. Create a merger for the shard-specific streams that uses the same order/grouping logic as the basic document provider

tokee:
[...]

PR #419, opened 2023-10-15
Harden fixed params for Solr requests

### LINKED ISSUE #418: Harden fixed Solr params
The property `solr.search.params` specieies parameters that must be applied to all Solr searches. Currently this is done by calling `SolrUtil.setSolrParams(solrQuery)`. This happens 35 different places in the code! There are at least 3 problems with the current approach:

1. It is very error prone to have to remember calling the method before issuing a query
2. The exact semantics as to when the the parameters should be applied (before or after local parameters) are unclear
3. Specifying a filter query (`fq`) in the fixed parameters overrides local filter queries if applies after the local filters has been applied. Or might be ignored if applied before the local filters, depending on the specific code

SolrWayback should move the fixed parameters mechanism into the `SolrClient` instead (by extending the class) and special case filter queries so that they extend existing filter queries instead of replacing them.

tokee:
The existing handling of fixed Solr parameters (`solr.search.params`in the properties) is error prone as each separate `SolrClient`-using functionality needs to call `SolrUtils.setSolrParams`. This pull request 

* Moves the functionality into `RestrictedSolrClient` so it happens automatically for all calls
* Handles filter queries (`fl`) gracefully as they are merged with existing filter queries instead of replacing them
* Makes it possible to (optionally) use query other collections than the one specified in `solrwayback.properties`. This foreshadows implementation of #329 where Solr's alias mechanism makes it necessary

No unit tests as the embedded Solr server does not expose a Solr URL with collection.

This closes #418

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/solr/NetarchiveSolrClient.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/solr/NetarchiveSolrClient.java
@@ -198,7 +198,7 @@ public class NetarchiveSolrClient {
         solrQuery.add("facet.limit", "" + (facetLimit + 1)); // +1 because itself will be removed and is almost certain of resultset is self-linking
         solrQuery.addFilterQuery("crawl_date:[" + dateStart + " TO " + dateEnd + "]");
         solrQuery.add("fl","id");                                                                                                                                                                  // request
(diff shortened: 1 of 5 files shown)


tokee:
Needs the second variant of `restrict` implementation and a check that caching takes collection into consideration.

thomasegense (approved):
Good cleanups also.

PR #421, opened 2023-10-16
Change CSV export to use ISO-8601 for time and not quote numbers

### LINKED ISSUE #323: Export of wayback data should use ISO-8601
When the endpoint `/export/fields` is called with the field `wayback_date`, it is formatted using a simple `toString()` resulting in outputs such as `Mon Jan 25 09:42:07 CET 1999`. It would be more consistent and easier to sort & parse to use ISO-8601: `1999-01-25T08:42:07Z`.

Fixing would be a simple check for type (Java `Date`) and subsequent formatting.

tokee:
This might require a minor version bump and a warning in the changelog as changing an export format might break tools expecting the old format.

This closes #323

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/export/GenerateCSV.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/export/GenerateCSV.java
@@ -1,13 +1,16 @@
 package dk.kb.netarchivesuite.solrwayback.export;
 
 
+import java.util.Date;
 import java.util.List;
 
+import dk.kb.netarchivesuite.solrwayback.util.DateUtils;
 import org.apache.solr.common.SolrDocument;
 
 /**
  * Created by teg on 10/28/16.
  */
+// TODO: This would be better with a dedicated CSV writer to handle the different types and escaping properly
 public class GenerateCSV {
 
     private static String NEWLINE="\n";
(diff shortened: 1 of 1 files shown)


src/main/java/dk/kb/netarchivesuite/solrwayback/export/GenerateCSV.java
```diff
@@ -72,8 +75,15 @@ public  static void generateLine(StringBuffer buffer,SolrDocument doc, String[]
                 if (field_value instanceof List) { //if multivalued
                     field_value = String.join(MULTIVALUE_SEPARATOR, (List<String>) field_value);
                 }
-                String escaped = escapeQuotes(field_value.toString());
-                result.append(escaped);
+                if (field_value instanceof String) {
```
thomasegense, line 78: Why not also "else if" here also for clarity? (and very minor optimization)

thomasegense (changes requested):

src/main/java/dk/kb/netarchivesuite/solrwayback/export/GenerateCSV.java
tokee: Fair point. I have made an unbroken `if-else` chain and adjusted the code so that it follows the same style in each of the code snippets

thomasegense (approved):

PR #422, opened 2023-10-17
RestrictedSolrclient extended to handle QueryRequests

tokee:
The pull request #419 contained a critical error where `QueryRequest`s were not supported. As these are used for plain search and faceting, basically nothing works.

This pull request adds param-adjustment to  `QueryRequest`-based requests.


thomasegense (approved):
Jeg deployer og tester på devel/pligt.

PR #437, opened 2023-11-23
no need to gzip for url replaced html

thomasegense:
Removed re-gzipping after HTML url  replace. There is no reason so do it.  It will fix a bug where gzip was not set
as correct content-encoding. It could have been set easy with a fix, but I prefer totally remove the gzip and making code simpler also,. Now playback html is always content-type:identify. (also for twitter, css etc.)


thomasegense:
Tested before and after on internal test server on multiple urls. Fixes playback of browsertrix harvest WARC's  and does not breake playback of sites that already work.

PR #438, opened 2023-12-20
warc_indexer now processes .arc-files too

fedorw:
the warc_indexer script now findes,processes and indexes .arc, .arc.gz, .warc and .warc.gz files.


tokee (approved):
Major oversight from my side. Thank you for catching!

Issue #440, opened 2023-12-23
Update solr dependencies in pom from 9.1.0 to 9.4.0


thomasegense:
The solrj 9.1 can still call a solr 9.4, so SolrWayback 5.0.0 is working as expected using the 9.1 dependencies.

But upgrading to 9.4.0 will break the unit tests that create and embeddedsolrserver.

java.lang.IllegalStateException: The resource configuration is not modifiable in this context.
	at org.glassfish.jersey.server.ResourceConfig$ImmutableState.register(ResourceConfig.java:229)
	at org.glassfish.jersey.server.ResourceConfig$ImmutableState.register(ResourceConfig.java:196)

I was not able to patch up the code. Googling did not help at all and only showed old code that does not work anymore.

Unittests:
EmbeddedSolrTest
UrlResolveTest

PR #441, opened 2024-01-05
Export speedup

### LINKED ISSUE #329: Improve export speed

tokee:
Major refactor of SolrWayback export. See the blogpost [Beware the cursorMark, my son!](https://sbdevel.wordpress.com/2023/10/24/beware-the-cursormark-my-son/) for overall principle.

The primary purpose of this pull request is to lower overall processing needs and increase performance when exporting from a multi-shard Solr setup. Single shard (aka Standalone) setups will see no change in speed. The easiest way to experiment with the alleged performance improvement is through the "unit test" `ShardStreamingTest.testStageSpeed` by ignoring the method name and pointing it to the production index :smiley: Do take care to not run it while a regular export is underway.

Tweaks to the improved export is done in `solrwayback.properties`. The sample setup in the `test`-folder holds the 4 properties, together with comments explaining what they do.

This closes #329 

Rewriting the export code gave a few secondary improvements:

* Core export code is now much cleaner, located in the `solr.SolrStream*` classes
* Throughput speed is logged during export every 10 seconds, instead of logging by fixed count
* Export `maxresults` limit is now checked at the last part of the export chain, making the limit reliable both with regard to deduplication and resource expansion

Note: Export speedup was implemented by Toke Eskildsen as an outside collaborator, **not** as part of his job at the Royal Danish Library.


thomasegense (changes requested):
Very solid work. JDK code quality one some classes like  CollectionUtils.
Not easy to understand code, but ease of usage is what is important, and framework is hidding it away.
I did a few very trivial fixes, so pull first.

**1:**
Is it correct that the export-stream does not set facet + highlight to false?
By luck facet=false iby default in edixmax definition, but I would be pleased if this was
made explicit by the code. But highlight is default 'on', so here it is important to turn it off. (even more speedup!!!)


**2:**
 Too excessive log.info when streaming?
Try run unittest and see output. This class in particular:
 StreamingSolrWarcExportBufferedInputStream
Can some be turned into debug?
And why is it logged double ? (same millis)

`11:49:30.189 [main] INFO  dk.kb.netarchivesuite.solrwayback.export.StreamingSolrWarcExportBufferedInputStream - Warcs read:270 arcs read:0
11:49:30.189 [main] INFO  dk.kb.netarchivesuite.solrwayback.export.StreamingSolrWarcExportBufferedInputStream - loadMore(): No more documents available after 270 docs read
11:49:30.189 [main] INFO  dk.kb.netarchivesuite.solrwayback.export.StreamingSolrWarcExportBufferedInputStream - warcExport buffer empty
11:49:30.189 [main] INFO  dk.kb.netarchivesuite.solrwayback.export.StreamingSolrWarcExportBufferedInputStream - Warcs read:270 arcs read:0
`
I will update the CHANGES.md :)

There is a few spin-offs from  this PR. But I want this PR merged fast due to 5.0 release and they most be solved later.
I will accept PR when the facet+highlight is fixed.

New issues herel isted here by my priority:
https://github.com/netarchivesuite/solrwayback/issues/443
https://github.com/netarchivesuite/solrwayback/issues/440
https://github.com/netarchivesuite/solrwayback/issues/442

tokee:
Your changes are causing a 100% diff and for the properties it seems that you are adding "windows linebreaks" (`CRLF`) instead of just `LF`!? It can be seen in GitHubs commit view.

thomasegense:
I hate eclipse more and more. Using it from my Windows computer decided to do this.
You welcome to overwrite it totally.

tokee:
Regarding facet+group+highlight: They are disabled in `SRequest.getMergedSolrQuery`. But thanks for the sanity check.
I have reduced logging somewhat. Feel free to reduce further.

I have no idea about the double logging and right now I don't have time for a deep dive into that. Sorry.

thomasegense:
And merged

PR #450, opened 2024-04-30
Change to GraalVM instead og Nashorn

VictorHarbo:
Building solrwayback with java versions above 15 fails the test class `JavascriptTests`.
This can be mitigated by changing which JavaScript engine is used to run these tests. Nashorn is discontinued in java 15+ and most services are changing to use GraalVM instead. This PR removes Nashorn and changes to GraalVM, which makes it possible to pass these tests on Java versions before and after 15. I've tested on 11 and 19.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,13 +162,6 @@
       <version>0.1.2</version>
     </dependency>
 
-    <!-- JavaScript testing -->
-    <dependency>
-      <groupId>org.openjdk.nashorn</groupId>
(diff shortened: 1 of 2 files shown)


VictorHarbo:
Converted to draft as I found a mistake. This update will make the porject not build on java 11 as it is right now.

VictorHarbo:
I've updated the first comment of this PR as it is now ready. We are not including Nashorn in the POM. I've changed to use GraalVM instead.

PR #458, opened 2024-07-24
Upgraded HttpSolrClient to Http2SolrClient

VictorHarbo:
Switching to the recommended Http2SolrClient and removing some un used imports found along the way. 

Closes #443

PR #459, opened 2024-08-01
Upgrade solr dependencies to 9.4.1

### LINKED ISSUE #440: Update solr dependencies in pom from 9.1.0 to 9.4.0

VictorHarbo:
Closes #440 

I've managed to upgrade the tests to not fail on 9.4 by upgrading lucene accordingly to: [solr 9 release notes](https://cwiki.apache.org/confluence/display/SOLR/ReleaseNote9_4_0) and changing how the embeded server is started.

I hoped that it would be possible to switch directly to 9.6.0 to also upgrade the deprecated HttpClient in #458, however there might be some greater obstacles going beyond solr 9.4. I'll try and look into that at some point.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # SolrWayback changelog
 
+UNRELEASED
+-----
+Upgraded solr dependencies from v9.1.0 to v9.4.1
 
 
 5.1.2
--- a/pom.xml
+++ b/pom.xml
@@ -113,20 +113,20 @@
       <dependency>
         <groupId>org.apache.solr</groupId>
         <artifactId>solr-solrj</artifactId>
-        <version>9.1.0</version>
+        <version>9.4.1</version>
       </dependency>
 
       <dependency>
         <groupId>org.apache.solr</groupId>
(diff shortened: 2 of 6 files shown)


thomasegense (commented):
Thanks. Now we also know to upgrade this is other projects as well.
I deployed on TEST environment and everything still seems to work the new upgraded SolrJClient from 9.4.1

PR #464, opened 2024-12-16
Fix gephi export limit

thomasegense:
Gephi export limit fixed.
Ascii logo added in console log at successful startup


VictorHarbo (approved):
Looks good

PR #465, opened 2024-12-17
Replace development jetty run with tomcat cargo run

thomasegense:
Try see if you can start development SolrWayback with new target and see log fil is working.


VictorHarbo (approved):
Works and looks good

Issue #463, opened 2024-10-07
Add download-button/link for Ngram-CSV


Klindten:
There used to be a download-button for Ngram search results.

Instead of correcting the q-parameter to force download - make a download - button/link.
[https://solrwb.kb.dk:4000/solrwayback/services/frontend/smurf/text/?q="thomas](https://solrwb.kb.dk:4000/solrwayback/services/frontend/smurf/text/?q=%22thomas) egense"&startdate=1998-01-01&enddate=2025-01-01&scale=YEAR

thomasegense:
@Klindten  You can test at: https://solrwb-test.kb.dk:4000/solrwayback/

PR #466, opened 2024-12-17
SOLR-463-should add a button for ngram to work

jesperlauridsen:
Errrh, I'm not quuuite sure if this is what was refered in the https://github.com/netarchivesuite/solrwayback/issues/463 issue, _but_ if not, then it's a great place to start. We had some export functionality already build, which exports into an CSV file. Is that whats needed here? Can you verify if thats what they mean, and check out if this code is what's needed then?


thomasegense (approved):
working!

PR #467, opened 2024-12-18
Make solr not return highlighting

### LINKED ISSUE #449: 'Visualize Result' Solr9 bug
**A dirty quick fix has been implemented in the v.5.1.0 release for this bug.**

When clicking the "Visualize Result" for a given search the parsing of JSON will fail in java code because Solr9 sends this attribute twice:
```
"highlighting":{},
"highlighting":{}`
```

And this is not a legal JSON object. Solr7 did not send any highlighting attributes.
A quick fix for release 5.1.0 is that I just do some ugly string manipulation of the JSON. A better way would be to avoid having Solr9 even returning this attribute that makes no sense for json.facet request.

The hack is implemented in the class: DomainStatisticsForDomainParser

The JSON can be seen by using this curl against a solr9 and a solr7.
`curl -s -d 'q=sports&rows=0&json.facet={domains:{type:terms,field:domain,limit:100 facet:{years:{type:range,field:crawl_year,start:2023,end:2024,gap:1}}}}' 'http://localhost:8983/solr/netarchivebuilder/select' > sports.json`

VictorHarbo:
Fix which closes #449


thomasegense:
Thanks, will test and merge tomorrow

thomasegense (approved):
Thanks, this is the right solution. And highlights was not needed, so also an
optimization.

I hope Solr will return valid JSON in a future version.

PR #468, opened 2025-01-10
Fix memento nullpointer

VictorHarbo:
Nullpointers and non working pagination is fixed by this PR


thomasegense (approved):
Tested and working

PR #470, opened 2025-07-09
Documentation cleanup

VictorHarbo:
So, this seems huge. I ran IntelliJs "Remove unused imports" which is the reason for the 39 of 41 changed files. Sorry for the jumpscare. 

I've tried to become a bit more acquainted with the Facade class and found that a lot of these methods really need a javadoc, as this class should be every new contributers first step into the project. I've tried to do a bit of cleaning and have added JavaDocs to a lot of the facade methods, but not all of them. 

Would you please double check, that the descriptions that I have provided are infact what the described code does?

Thanks in advance and sorry for the many changed files.


thomasegense:
Thanks. I hope to do take a look on it tomorrow.

thomasegense (approved):
Thanks for the boring job of writing javadoc. The Facade class is the most important class to have good javadoc for in SolrWayback. So it was much needed.
Also thanks for small refactoring contains()+move method to util.

I can not delete the PR branch since it is in your repository. You have to clean that up yourself :)

PR #471, opened 2025-07-21
Remove port from url norm

thomasegense:
Very small fix that removes URL in url_norm field fixing playback bug with resources not found.
For LEGACY mode the port is not removed. The option is slightly better for very old solr index already build on solr 7 with old version of warc-indexder.


src/main/java/dk/kb/netarchivesuite/solrwayback/facade/Facade.java
thomasegense, line 1: This is a fix for regression error from last refactoring branch.

kesj-kb (approved):

VictorHarbo (approved):
Looks good

PR #480, opened 2025-11-08
GSON dependency update

VictorHarbo:
Update gson dependency version to 2.13.2. Only affects class `SolrStatsTest`.


thomasegense (approved):
LGTM

PR #481, opened 2025-11-08
Bumped commons-fileupload

VictorHarbo:
Sorry for individual small PRs.
Bumping commons.fileupload, only used in `SolrStreamDirectTest`, test still runs


thomasegense:
If it is only used from a unit test, the dependency should be marked as test-only resource in the dependency.
Check if it still compiles.

VictorHarbo:
Added test scope to the dependency.

thomasegense (approved):
LGTM

Issue #483, opened 2025-11-15
Improve test coverage


VictorHarbo:
# Overall test coverage:

When running the test suite with coverage in IntelliJ IDEA results are as follows:

| Class %       | Method %       | Line %          | Branch %       |
| ------------- | -------------- | --------------- | -------------- |
| 65% (120/184) | 44% (765/1731) | 41% (3126/7503) | 38% (881/2273) |

## This issue 

This issue is created to keep track of updates in test coverage as I am currently working on providing the project with better overall test coverage. 

This issue will be used as an overall tracker for the coming PRs. I will try to keep PRs at a somewhat reasonable size instead of combining all additions into one. This way updates are also easier to follow.

PR #484, opened 2025-11-15
Improve test coverage

VictorHarbo:
This is the first of hopefully many PRs implementing unit tests for untested parts of SolrWayback. 

Related to #483 

This moves test coverage to 809/1715 methods from 765/1715.

| Class %       | Method %       | Line %          | Branch %       |
| ------------- | -------------- | --------------- | -------------- |
| 68% (121/177) | 47% (809/1715) | 44% (3474/7723) | 38% (943/2443) |

### DIFF
--- a/src/main/java/dk/kb/netarchivesuite/solrwayback/export/StreamingRawZipExport.java
+++ b/src/main/java/dk/kb/netarchivesuite/solrwayback/export/StreamingRawZipExport.java
@@ -59,7 +59,7 @@ public class StreamingRawZipExport {
      * @param warcMetadata  Object to save metadata information to.
      * @return              The input SolrDocument is returned, which makes this method stream compliant.
      */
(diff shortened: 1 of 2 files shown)


thomasegense (changes requested):
Thank you.
Much needed unittests.
The unit tests spinning up an embedded Solr with data are perfect and tests the full circle (after the indexer has run).
I only have 1 request change and only for the  StreamingRawZipExportTest.
Using reflect to call methods that are private are problematic, I know this is only way
you can call them unless you can change the "private" modifier.

Here are some of the cons:

The must common used solution is change the method modifier from private to 'protected'..
Then it can be called from unittest classes that has the same package name as the class it is testing. So can still not be soon the class API for "outsiders".

Changing the methods to public will also fix the problem, but is also not good code practice, since they now can be called from everywhere and nobody excepts unittest should know about them.
Changing the method name will not give compile error because the method is not refereced directly in the unittest.

PR #487, opened 2026-01-09
Navigation Tracker

VictorHarbo:
This PR implements a Navigation Tracker feature which is highly relevant for academics using the software for exploring collections of archived web material.

The feature logs all navigation in the application and implements a button for downloading the history from a session. 

Implementation considerations:
- Navigation History is tracked server side to make clicks in playback trackable
- Navigation History is tracked by session to make all users have their own and only their own navigation available
- The frontend calls bypasses the Vue Store as we see this as a function that should be able to be turned on and off in the future

The PR has been reviewed and tested internally in the WEBCHILD project by @jorntx, his review can be seen here: https://github.com/WEB-CHILD/solrwayback/pull/1


thomasegense (approved):
Not time for deep review/testing. I trust you guys)
Testing will be done on the 5.4.0 candidate that will be marked as pre-release at first.

PR #488, opened 2026-01-20
Rollback change making datafields endpoint not work

VictorHarbo:
So, I wanted to make an endpoint return pretty and valid JSON. This broke the show datafields button in the frontend. This rolls back the change. 

You should probably release a patch version of this. Sorry for the inconvinience.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,20 +1,18 @@
 # SolrWayback changelog
 
 UNRELEASED
------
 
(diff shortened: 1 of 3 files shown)


thomasegense:
Since the release is out, I need to patch fast. I will make a 3.4.1 release.

For future PR we probably need to have @jesperlauridsen  review.

Issue #495, opened 2026-03-12
Changes to the NavigationHistory feature.


jesperlauridsen:
Hey @VictorHarbo / @jorntx / @thomasegense.

I didn't get to review the NavigationHistory, but I looked through it after the merge, and was puzzled by the way of handling storage of the history count. (I also still don't quite understand the feature, but that's on me ;-))

The way it currently works, as I understand it, is that you push changes to the backend, and then every 2 seconds asks the backend for a new object containing the count. This is done regardless of any chances made.

<img width="1241" height="647" alt="Image" src="https://github.com/user-attachments/assets/9c531677-5b37-44b5-b9e4-d64363f1effc" />

I would recommend that this should be rewritten so a store on the frontend keeps track of everything, and when a change is made, it's pushed to the backend while keeping track at the frontend aswell. We can then ensure that on a fresh load of the page / or a reload, we ask the backend for the count object, and load it into the store, and keep track of it there once again.

That way, we avoid asking for updates (and often just getting the same object with no changes) from the backend. The current appoarch doesn't really scale, imho, if we some day have an instance of solrwayback deployed somewhere with alot of concurrent users.

Also, I think the button should either be  renamed to something along the lines of "Download navigation history", or instead open a modal of the navigation history with the option to download it. Right now you just suddenly get a json file thrown into your download folder. :-)

VictorHarbo:
Hi @jesperlauridsen 

Nice to have your frontend experienced eyes looking at the feature! Thank you! 

First of all, the purpose of the feature is for users to be able to document their discovery pathways through the archive. In our project, this has already proven useful in multiple ways. First of all, having the navigation history available makes it possible for researchers to pick up their research where they left off during an earlier session. The navigation history also serves as documentation for how a specific source has been found in the archive, as one of the biggest challenges in such big collections is to keep track of how a specific resource has been found, as this is crucial for academic transparency and reproducibility.

Regarding the implementation, it probably shines through that I am a bit more comfortable handling things on the backend. Your store-based approach seems solid and is probably a good improvement. For me, it would make sense if the button opened a modal showing the history, but also still included the possibilities for download, for storing the history as documentation.

VictorHarbo:
Also, going forwards, how are your time prioritised at the Library. I have assigned Thomas as reviewer for everything as that has been how we've normally done it. But would love to assign you to frontend reviews if you are available for this?

thomasegense:
[...]

PR #500, opened 2026-06-03
Update Navigation History file name

VictorHarbo:
Our research team members are downloading a lot of their navigation history files and they wanted easier filenaming. 
I've implemented it as YYYYMMDD_navigation_history.json and @jorntx has reviewed internally. 

We will hopefully have some time during the summer or fall to work more on this feature, including creating the modal view and remove the polling completely as we have discussed earlier.

### DIFF
--- a/src/js/src/components/NavigationHistoryButton.vue
+++ b/src/js/src/components/NavigationHistoryButton.vue
@@ -93,12 +93,22 @@ export default {
         if (!response.ok) {
           throw new Error('Failed to download history')
         }
-        
+
+
+       // Parse filename header
+       const disposition = response.headers.get('content-disposition') || ''
+       let filename = 'query_history.json'     // fallback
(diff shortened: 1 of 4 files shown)


thomasegense:
It is too much to also remove the auto pull completely? On master I have increased it to 60 seconds (from 2 seconds).
New logic:
1) At first search notify the box there is some history. Drop the count.
When downloading you will still download the full history.

Next step (another PR):
1) Clicking history opens a modal with download option. (here user can change the new default name)
2) Option to reset history. (For public libraries etc. with walk-in. History has to be cleared).

VictorHarbo:
Hi Thomas! Yes, it should be removed completely in a future PR. 
Would it be alright for you if we keep it as it is here and then we rework it later as described here and in #495?

jesperlauridsen (approved):
Looks good for my part :-)

thomasegense (approved):
LGTM

PR #505, opened 2026-06-25
Ngram UI fixes, URL normalisation unit tests and dependency upgrades

jorntx:
## Ngram UI fixes, URL normalisation tests and dependency upgrades

This PR collects a set of bug fixes, dependency upgrades, and test additions accumulated since the last upstream merge.

---

#### Ngram chart / search box

Several interconnected bugs were present in the ngram search UI. They are fixed in this PR:

- **About text re-appearing on query clear** — the about/intro text was being shown again whenever the user manually deleted the text from the ngram search box, because the clear action was not being distinguished from a genuine "empty query" state. Fixed by tracking the clear action explicitly.
- **Chart not clearing on search-box clear** — invoking the clear function in the search box left the previous chart visible. A dedicated `clearChart` function is now called to tear down the existing chart instance when the query is cleared.
- **Click on data point did not trigger a search** — clicking a point on the ngram line chart was supposed to deep-link into a filtered search, but the handler was silently failing. The bug is now resolved so clicking a data point correctly initiates the search.
- **Layout shift / bounce during searches** — the search box was shifting position while a search was in flight, causing a noticeable layout jump. The layout is now stable throughout the search lifecycle.
- **`TypeError: undefined is not an object` during ngram searches** — a missing null-guard caused a hard crash under certain search conditions. Fixed with a defensive check before accessing the relevant object.
- **`TypeError: undefined is not an object` when exporting chart data** — the CSV export logic was referencing a Vue 2-era data shape that no longer exists. The export mixin has been refactored to read directly from `ngramStore` datasets and labels, removing the stale legacy code path.
- **Chart sizing and incomplete options** — the ngram chart was rendering at the wrong size, and only a partial options object was being passed to the chart engine. Both the sizing CSS and the options wiring are now corrected so the full configuration is applied on every render.

---

### Dependencies

Security/maintenance upgrades to keep the build toolchain current:

| Package | Old | New |
|---|---|---|
| `solr` (backend) | `9.6.0` | `9.10.1` |
| `vite` (frontend) | `6.3.5` | `^6.4.2` |
| `vitest` (frontend) | `3.1.3` | `^3.2.6` |

Both `vite` and `vitest` bumps address known CVEs in the prior versions.

---

### Tests

New and improved test coverage for the URL normalisation layer:

- **`NormalisationStandard.canonicaliseURL`** — expanded unit tests covering the canonical URL construction logic.
- **`PathResolver.waybackAPIResolverHelper`** — new test class (`PathResolverWaybackAPIResolverTest`) with broad coverage of the wayback API resolver helper, including port-stripping behaviour introduced in this PR.
- **`UrlUtils` url\_norm query building** — tests for `url_norm` query construction, including default-port handling (port 80 stripping) to prevent regressions on the fix above.

---

**Files changed:** 16 files, +503 / −88 lines  
**Affected areas:** ngram frontend components, CSV export mixin, Solr client, test suite

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,7 +7,7 @@
     "**/.factorypath": true
   },
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "eslint.validate": ["javascript", "vue"],
   "eslint.workingDirectories": [
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,7 +7,7 @@
     "**/.factorypath": true
   },
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": "explicit"
+    "source.fixAll.eslint": true
   },
   "eslint.validate": ["javascript", "vue"],
   "eslint.workingDirectories": [
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,9 +10,19 @@
 * Switched from deprecated NoOpResponseParser to the new InputStreamResponseParser. Caching of SolrResponses is now handled by storing the response in the NetarchiveSolrClient.
 
 ### Fixed
+* About text no longer reappears when manually clearing the ngram search query
+* Ngram chart is now properly cleared when the search box clear function is invoked
+* Fixed bug preventing searches from being performed when clicking on data points on ngram chart
+* Resolved bouncing/layout shift of the search box during ngram searches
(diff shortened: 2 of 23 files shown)


jesperlauridsen (approved):
A quick look over the code gave me no worries. I haven't spun it up to check it - but it looks solid enough, imho.

@thomasegense, if we need to test it, can we put it on a devel instance somwhere?

thomasegense:
I have tested it.

thomasegense (approved):
It is working.
Nice changelog details.

PR #509, opened 2026-09-04
Extend Solr docs for separate installs, allowPaths, and collection-id #460

### LINKED ISSUE #460: Extension of Solr documentation
Proposal for some updates and extensions for Solr documentation:

1. Reference [Extended Solr guide](https://github.com/netarchivesuite/solrwayback/blob/master/src/bundle/solr/SOLR.md) already in the main README file, just for the case, when user wants to install Solr separately. The [part](https://github.com/netarchivesuite/solrwayback/blob/ba4d691e75bde3ef926947810a71a6ccb4b3c2f3/src/bundle/solr/SOLR.md?plain=1#L46) where Solr config from webarchive-discovery is used to create Solr Cloud is very important.
2. Since Solr-9 it is not possible to store index files outside of Solr home folder (e.g. for the case, when external file storage is used), some additional setting of solr.allowPaths is needed for that, see [here](https://stackoverflow.com/questions/76169675/how-to-set-solr-index-data-path-version-9-2-0). This note should be mentioned in Extended Solr guide.
3. Usage of "--collection_id" in custom configuration of the indexer in the main README:
https://github.com/netarchivesuite/solrwayback/blob/ba4d691e75bde3ef926947810a71a6ccb4b3c2f3/README.md?plain=1#L228
Currently the spelling of "collection_id" was changed to "collection-id", see issue: https://github.com/ukwa/webarchive-discovery/issues/315 . So, maybe with new release it will work again with "collection_id".
Also for custom configuration the parameter "--institution institution1" also could be added as an example to the README file, to have all 3 possibilities. I think, we have only 3 options here: https://github.com/ukwa/webarchive-discovery/blob/13595bead029fd44f133ec6c18f689edde202e53/warc-indexer/src/main/java/uk/bl/wa/solr/SolrFields.java#L151-L154

AzazelSensei:
I added the three doc updates from #460.

The README now points at the Extended Solr guide for a separate Solr 9 Cloud install that uses the webarchive-discovery config. SOLR.md notes `solr.allowPaths` when the index is outside Solr home. The custom indexing example uses `--collection-id` and `--institution`.

Closes #460

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,13 +133,16 @@ Absolute URL live-leaks (starting with http://domain...) will not be caught and 
 #### Warc indexer
 Documents in SolrWayback are indexed through the [warc-indexer](https://github.com/ukwa/webarchive-discovery/tree/master/warc-indexer), which is maintained by The British Library. This is a fundamental tool for indexing WARC files in SolrWayback and furthermore it is included in the [SolrWayback bundle](https://github.com/netarchivesuite/solrwayback/releases).
 
+#### Extended Solr guide
+If you install Solr separately instead of using the Solr that ships in the bundle, see the [Extended Solr guide](src/bundle/solr/SOLR.md). That guide covers Solr 9 Cloud, including uploading the webarchive-discovery Solr configuration and creating the `netarchivebuilder` collection.
+
  
 ## Requirements
  * Works on macOS/Linux/Windows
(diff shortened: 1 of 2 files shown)


thomasegense (approved):
LGTM

Issue #510, opened 2026-09-07
Add host-level granularity to the Wordcloud, Domain Stats and Link Graph features in the Toolbox


bnfleb:
In order to analyse the archive statistics in greater detail, we would like to be able to use the wordcloud, domain stats and link graph features with host-lvele granularity in the toolbox.

To this end, here are the changes that have been made: 

- For Wordcloud, Domain Stats and Link Graph: possibility to choose between domain or host. The default value is domain.
- Domain stats: renamed to Stats
- Linkgraph: if the user choose to use the host granularity, the query will look at the host and links_host fields in Solr.

thomasegense:
Merged into master. New SolrWayback bundle release:
https://github.com/netarchivesuite/solrwayback/releases/tag/5.5.0

PR #511, opened 2026-09-07
Host-level granularity to the Toolbox's WordCloud, Stats and LinkGraph

bnfleb:
Linked to #510.

For the wordcloud : I saw that the method wordCloudForDomain will be deleted when the frontend will change. In the meantime, I change it in "wordCloud" and it generates the image for a domain or an host.

For the domain stats : I renamed the tab from Domain stats to Stats and a few methods but I didn't change the DomainStatistics object, the DomainStats.vue and all the associated css. Should I harmonize the code ?

### DIFF
--- a/src/js/src/assets/styles/search.scss
+++ b/src/js/src/assets/styles/search.scss
@@ -452,7 +452,7 @@ color:var(--secondary-text-color);
 }
 
 .wordcloudContainer .wordcloudButton {
-  margin-top: 15px !important;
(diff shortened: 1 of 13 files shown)


thomasegense (approved):
I have tested at it seems to work as expected.
No need to rename method/DTO to match domain/host better.

jesperlauridsen (approved):
Looks good frontend-wise. Can't speak for the changed backend services :-)

